PR #15, opened 2018-09-24
Authentication

patricksimonian:


patricksimonian:
This is the first PR in prep for development of the auth mechanism. Just added some ui components for things like 'login' buttons

PR #33, opened 2018-10-09
Integrate authentication mechanics

patricksimonian:
- sso base url now arriving from environment variable
- fixed some broken unit tests

### DIFF
--- a/app-web/gatsby-config.js
+++ b/app-web/gatsby-config.js
@@ -1,4 +1,7 @@
-require('dotenv').config();
+require('dotenv').config({
+  path: '.env.production',
+});
+
 module.exports = {
   siteMetadata: {
     title: 'Devhub',
--- a/app-web/src/components/PrimaryHeader/PrimaryHeader.js
+++ b/app-web/src/components/PrimaryHeader/PrimaryHeader.js
@@ -1,59 +1,17 @@
 import React from 'react';
-import PropTypes from 'prop-types';
-import { connect } from 'react-redux';
-
 import classes from './PrimaryHeader.module.css';
 
 import { APP_TITLE } from '../../constants/strings';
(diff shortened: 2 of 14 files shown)


app-web/gatsby-config.js
```diff
@@ -1,4 +1,7 @@
-require('dotenv').config();
+require('dotenv').config({
```
jleach: I think dev and test will be the only time we use a `.env` file. The rest of the time it will come from ENV vars. Consider wrapping this in a `if (process.env.NODE_ENV === 'test' || process.env.NODE_ENV === 'development'). In a few projects we have used `nconf` to help manage and merge sensitive and non-sensitive config into a `config` object that can be used like `config.get('sso:clientId')` see https://github.com/bcgov/range-api/tree/develop/server/config

app-web/src/components/PrimaryHeader/PrimaryHeader.js
```diff
@@ -1,60 +1,17 @@
...
-        clicked={() => {
-          implicitAuthManager.clearAuthLocalStorage();
-          window.location = implicitAuthManager.getSSOLogoutURI();
-        }}
-      >
-        Logout
-      </Button>
-    );
-  }
+import Login from '../Auth/Login/Login';
```
jleach: Personally I think its best to stick with lower case paths were possible. They can be finicky in different environments: Mac does no care; Linux cares; Windows does not care.

app-web/gatsby-config.js
patricksimonian: I like that approach much better. I'll set up an issue for it to keep track of things. Gatsby has it's own way of passing sensitive/non sensitive data to and from the configuration process to the static site building. So I'll have to see if it's possible to work it in there

PR #43, opened 2018-10-16
Local Gatsby Source Plugin for retrieving files from repos

patricksimonian:
Several file changes are just formatting changes from Prettier js


app-web/plugins/gatsby-source-github-all/utils/constants.js
```diff
@@ -0,0 +1,54 @@
...
+
+const MEDIATYPES = {
+  md: 'text/markdown',
+  json: 'application/json',
+  yaml: 'text/yaml',
+  txt: 'text/html',
+  html: 'text/html',
+};
+
+const PROCESSABLE_EXTENSIONS = ['.md'];
```
ShellyXueHan: do you need to export it?

app-web/plugins/gatsby-source-github-all/utils/github-api.js
```diff
@@ -0,0 +1,223 @@
...
+ * b64 encoded string of the file contents into either raw string data or json
+ * @param {String} repo 
+ * @param {String} owner 
+ * @param {String} path 
+ * @param {String} token 
+ */
+const fetchFile = async (repo, owner, path, token) => {
+  try {
+    const result = await fetch(
+      `https://api.github.com/repos/${owner}/${repo}/contents/${path}`,
```
ShellyXueHan: use constant

patricksimonian: switched out for constant : )

app-web/plugins/gatsby-source-github-all/utils/constants.js
patricksimonian: yes i do. Good catch!

PR #54, opened 2018-10-23
Imp/remark frontmatter defaults

patricksimonian:
add a structure for providing defaults to files when ingested by gatsby


app-web/plugins/gatsby-source-github-all/utils/transformer.js
```diff
@@ -0,0 +1,53 @@
...
+ * takes in file parameters and runs it through plugins that
+ * modify the content
+ * @param {String} fileExtension 
+ * @param {String} content 
+ * @param {Object} file 
+ * @returns {String} file content transformed
+ */
+const fileTransformer = (fileExtension, content, file) => ({
+    content,
+    use(plugin, options = {}) {
```
jleach: Where does use come from?

app-web/plugins/gatsby-source-github-all/utils/plugins.js
```diff
@@ -0,0 +1,77 @@
...
+        });
+        // create 'new' md string with updated front matter
+        return matter.stringify(data.content, frontmatter);
+    }
+    return raw;
+};
+
+module.exports = {
+    markdownPlugin,
+};
```
jleach: You get these little icons at the bottom of files if they don't have new lines. Generally I think its best practice to have one at the end of a file.

app-web/plugins/gatsby-source-github-all/utils/transformer.js
patricksimonian: i think think this is a es6 function definition, similar to 
```javascript
{ 
  content,
  use: function(plugin, options = {}) {...},
}
```

patricksimonian: i may be wrong on its being an es6 definition, it was eslint that actually auto corrected it to what you see now

app-web/plugins/gatsby-source-github-all/utils/plugins.js
patricksimonian: gotchya. corrected :)

PR #58, opened 2018-10-25
Imp/markdown path transform plugin

patricksimonian:
# Path Transform Plugin
Created a markdown remark plugin to sift through a markdown ast tree and call a cb on any images or links that are using relative paths. 
It is the Cb's job to create a absolutepath from the relative path (given the cb arguments)

In addition I made some tweaks the the gatsby github source all function. Added some refactoring. The fetch file routine now returns undefined on failed fetches (responses other than 200s) These failed fetches are filtered out before creating graphql nodes to prevent the build failing because a single file failed to fetch. Gatsby will just build without that file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ In Development
 
 - [Devhub Ingestion System](./docs/gatsbySources.md)
 
+- [Authoring Plugins for Remark](https://www.huynguyen.io/2018-05-remark-gatsby-plugin-part-2/)
 ## Security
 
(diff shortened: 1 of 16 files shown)


patricksimonian:
You may notice a npm ignore file in the remark transformer plugin I wrote. That is because this plugin is being prepped for eventual publishing to npm. I have the package json for that plugin prepped with dependancies minus the lock file as I'm not using node modules from the plugin but rather from the global project directory at this stage.

PR #69, opened 2018-10-31
import source nodes static links

patricksimonian:


app-web/plugins/gatsby-source-github-all/utils/plugins.js
```diff
@@ -15,7 +15,7 @@ limitations under the License.
 
 Created by Patrick Simonian
 */
-
+const shortid = require('shortid'); // eslint-disable-line
```
jleach: I think you can wrap all of these in` /* eslint-disable */` and `/* eslint-enable */`.

patricksimonian: i'll check that out

PR #72, opened 2018-11-01
Update issue templates

patricksimonian:


patricksimonian:
added some issue templates, thoughts on the templates? we can change their content to suit our needs

PR #87, opened 2018-11-08
Bug/path to resource

patricksimonian:
## Issue
Resource Paths were unstable and changed values between builds. 
## Affects
For users, bookmarks to resources would break after a new build.

## Solution
paths to resources are now stable. They do not change values between builds.
For example:
Siphon sources `footer.md` from the repo `design-system`

Originally the resource path would `/design-system/footer_[random string]`

**Now** a hash is generated from `source + filename // ie 'design-system' + 'footer.md'` so the resource path becomes
`/design-system/[stable-unchanging-hash]`

##

### DIFF
--- a/app-web/plugins/gatsby-source-github-all/__tests__/helpers.itest.js
+++ b/app-web/plugins/gatsby-source-github-all/__tests__/helpers.itest.js
@@ -1,0 +1,14 @@
+import { createPathWithDigest } from '../utils/helpers';
+jest.unmock('@bcgov/common-web-utils');
+jest.unmock('shorthash');
+const shorthash = require('shorthash');
+describe('createPathWithDigest Integration Tests', () => {
+  it('return path is idempotent', () => {
+    const digestable = 'readme.md';
+    const hash = shorthash.unique(digestable);
+    expect(createPathWithDigest('/path/', digestable)).toBe(`/path/${hash}`);
+    expect(createPathWithDigest('path/', digestable)).toBe(
+      createPathWithDigest('path/', digestable)
+    );
+  });
+});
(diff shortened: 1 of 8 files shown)


app-web/plugins/gatsby-source-github-all/__tests__/helpers.itest.js
```diff
@@ -0,0 +1,14 @@
+import { createPathWithDigest } from '../utils/helpers';
```
jleach: A bit hard to read when everything is bunched up like this; a little spacing would be ok.

app-web/plugins/gatsby-source-github-all/utils/plugins.js
```diff
@@ -95,8 +96,10 @@ const markdownPagePathPlugin = (extension, file) => {
   } else {
     // no resource path, this file is destined to be turned into a page,
     // the page page is composed of the source name, the title of the file plus an id
-    file.metadata.resourcePath = `/${file.metadata.source}/${file.metadata
-      .name}_${shortid.generate()}`;
+    file.metadata.resourcePath = createPathWithDigest(
```
jleach: You could probably just pass in `file.metadata.source` and `file.metadata.name` as the two params then `+` them as needed in the function.

app-web/plugins/gatsby-source-github-all/utils/helpers.js
```diff
@@ -0,0 +1,42 @@
...
+ * returns an idempotent path based on a base path plus a digestable string that is hashed
+ * @param {String} source the string name of the original source
+ * @param  {Array} digestable 
+ * @returns {String} ie (/mypath, file.md) => /mypath/123dsfakjhdf
+ */
+const { TypeCheck } = require('@bcgov/common-web-utils'); // eslint-disable-line
+const path = require('path');
+const shorthash = require('shorthash');
+
+const createPathWithDigest = (base, digestable) => {
```
jleach: A bit bunched up again. To make it more readable I'd consider a space before and after the 2nd `if` then one before the return.

app-web/plugins/gatsby-source-github-all/utils/plugins.js
patricksimonian: updated the function to take a spread of strings (basePath, digestable1, digestable2 ...etc). The spread of strings are joined.

PR #98, opened 2018-11-16
Add support for 'resourceTypes' for Siphon Nodes + updated documentation for Siphon

patricksimonian:
This is an incremental pull request for the source filtering feature


patricksimonian:
please ignore all Toggle related files, these were pulled in from sprint-5 because of some dependancies I required but otherwise have already been merged and reviewed

patricksimonian:
this was approved by shelly

jleach:
@patricksimonian Consider adding more white space to make the code easier to read; were a few file in here that read like a WoT (Wall of Text). While I think code documentaiton is great and I'm not the best and adding them all the time, I think they're better used to document stuff that's not obvious or anothe way, make sure they're not superfluous. For example, there shouldn't be any need to docuemet what `_.flatMap()` is doing. And finally, that was a pretty big PR. You'll generally get no good feedback on large PRs. Try and keep them under 300 lines of changed code.

```
    // fetch ignore file if exists
    const repoIgnores = await fetchIgnoreFile(repo, owner, token, branch);
```

patricksimonian:
thanks for all of that feedback. I'll definitely try to make the PR's smaller than this one. Also really good point on code readability. I'll take that into account from now on

PR #106, opened 2018-11-19
Feat/unfurl external links

patricksimonian:
# About
External Unfurl Plugin transforms files where the `resourcePath` points to an external link. The path is unfurled and any meta data is scraped.

## Changes
- added packages necessary for plugin
- created plugin


app-web/plugins/gatsby-source-github-all/utils/plugins.js
```diff
@@ -164,9 +177,26 @@ const markdownResourceTypePlugin = (extension, file) => {
...
+ * @param {String} extension 
+ * @param {String} file 
+ * @returns the modified file
+ */
+const externalLinkUnfurlPlugin = async (extension, file) => {
+  // does file have a resource path and is it a valid url?
+  if (file.metadata.resourcePath && validUrl.isUri(file.metadata.resourcePath)) {
+    const { body: html, url } = await got(file.resourcePath);
+    const metadata = await metascraper({ html, url });
+    file.metadata.unfurl = createUnfurlObj('external', metadata);
```
jleach: Keywords like 'external' might go well in `constants.js`. That way if you need to change it later you can just edit one place rather than search-replace in the code base. Only change this if it makes sense (makes the code more maintainable).

```js
export const LINK_TYPE = {
  EXTERNAL: 'external',
  INTERNAL: 'internal',
};```

patricksimonian: good point, i'll make that change !

patricksimonian:
had to force a push, i mistakenly included some commits that i didn't want in this pr

PR #107, opened 2018-11-19
Implement Async Transformer and Update/Add Tests

patricksimonian:
# About
As per #103, this PR contains those changes + updates to unit/integration tests.

## Changes
- updated FileTransformer to run asyncronously
- updated all unit/integration tests to respect FileTransformer Changes
- Added unit tests for external unfurl plugin
- Attempted at improving the readability for some of the code

### DIFF
--- a/app-web/__mocks__/got.js
+++ b/app-web/__mocks__/got.js
@@ -1,0 +1,6 @@
+jest.requireActual('got');
+
+module.exports = async url => ({
+  url,
+  body: null,
+});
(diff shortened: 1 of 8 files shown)


app-web/__mocks__/got.js
```diff
@@ -0,0 +1,6 @@
+jest.requireActual('got');
```
jleach: I think you need all your files to have the standard disclaimer blerb at the top. I don't recall the history but I was advised to do it by Todd when I started and he's quite knolidable about licencing best practices.

jleach:
I think you need all your files to have the standard disclaimer blerb at the top. I don't recall the history but I was advised to do it by Todd when I started and he's quite knolidable about licencing best practices.

PR #112, opened 2018-11-20
Add dynamic page creation to gatsby configuration

patricksimonian:
**Please note** these changes are very specific to gatsby v1 and ***will*** break on upgrade to gatsby v2
# About
Another step in the dynamic filtering feature. Filtering will be done via 'restful routes'. 
IE navigating to `/resources/components` will just re render the index page to show only cards that are `resourceType === 'components'`

During the navigation process, gatsby will actually look for a static site at the path `/resources/components`. The following changes tells gatsby to treat this route as dynamic and client side rendered only. 

# Changes
- added a hook for `onCreatePage` to treat any routes `/resources/*` as client side rendered
- did some code cleanup in the `gatsby-node.js`


patricksimonian:
incorrectly pulled into master

PR #116, opened 2018-11-21
Filter Feature refactored to use Redux store

patricksimonian:
# About

To give more transparency to how the filter feature works, it was converted from a locally state managed component to a globally managed redux reducer. 

This also allows us to use the state of the filter feature in other components. IE we could render our a header saying 'Components' somewhere on the page if the filter option had selected to filter Components

## Changes
- changed state to redux store
- renamed component from `<PrimaryNavigation />` to `<PrimaryFilter />`
- refactored to `<PrimaryFilter />` functional component


app-web/src/components/PrimaryFilter/PrimaryFilter.js
```diff
@@ -0,0 +1,140 @@
+/*
```
patricksimonian: Only change made to this file was  it was converted from a `class Component` to `const Component`

app-web/src/components/PrimaryFilter/PrimaryFilter.js
```diff
@@ -0,0 +1,140 @@
...
+        onMenuClose={() => setMainNavigationToggled(false)}
+      />
+      {/* mobile select */}
+      <Hamburger
+        className={styles.mobileOnly}
+        clicked={() => setMainNavigationToggled(!menuToggled)}
+      />
+      <Select
+        value={selectedOption}
+        ariaLabel={ARIA_LABEL_FILTER_SELECT}
```
jleach: Area?

PR #124, opened 2018-11-22
task-430 Designed and implemented a 404 page in React and Gatsby. 

rajpersram:
Created the React components for the 404 page involving: 404 calling venndiagram which then calls circles. If there's anything that can be better written for React, please let me know. Thanks!

### DIFF
--- a/app-web/src/components/UI/VennDiagram/VennDiagram.js
+++ b/app-web/src/components/UI/VennDiagram/VennDiagram.js
@@ -1,0 +1,23 @@
+import React from 'react';
+import classes from './VennDiagram.module.css';
+import Circle from '../Circle/Circle';
+
+
+const VennDiagram = ({ location }) => (
+    <div className={classes.venn_diagram}>
+        <div>
+            <Circle type="circle-first"><p>We missed something.</p></Circle>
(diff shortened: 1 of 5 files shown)


app-web/src/components/UI/VennDiagram/VennDiagram.js
```diff
@@ -0,0 +1,23 @@
...
+
+const VennDiagram = ({ location }) => (
+    <div className={classes.venn_diagram}>
+        <div>
+            <Circle type="circle-first"><p>We missed something.</p></Circle>
+            <Circle type="circle-second"><p>You mistyped.</p></Circle> 
+            <p className={classes.center}>404 Error</p>
+        </div>
+        <div>
+            <a href="">Take me home.</a>
```
patricksimonian: Look really good! 
In the spirit of a reusable component, I think the lower half of this code could probably just live in the 404.js page. We will take a look at refactoring this component further to make it fully customizable. The options are endless : D

rajpersram: Thanks. The trick was having the 'take me home' bit to show in the middle of the markup of the Venn diagram. Instead, I just put it at the top before the diagram is called.

PR #125, opened 2018-11-22
remove ignore param in siphon

patricksimonian:


patricksimonian:
closing as this is already addressed in sprint-5

PR #131, opened 2018-11-23
Fix the `<img>` tags not working in md files issue 

patricksimonian:
# About
Markdown files support the use of some html within the content. Things like 
```md
![Source](../../source.png)
<img src="../../source.png" alt="Source" >
```
Should be equivalent. 

There was an issue where using the `img` tag in a markdown file would show up broken in the devhub if
the `src` attribute was a relative path. 

This was fixed by modifying the *transformer* plugin responsible for fixing relative path issues #58 to also 
look into html ast nodes within the markdown file and transform any image tags as needed.

## Changes
- refactored and updated the remark-transformer-path plugin to manipulate html ast nodes that have img tags within them
- updated unit tests


patricksimonian:
closing and reopening

PR #133, opened 2018-11-23
add auto link headers

patricksimonian:
# About
Internal Navigation for a md document wasn't working in devhub

## Changes
- this has been fixed by adding the auto link headers transformer plugin

*** IMPORTANT ***
The version of the npm auto link headers package has been purposely set to `1.4.19` as that is the last stable version that is compatible with Gatsby V1, in reality this plugin is already in 2.x.x and so we will need to upgrade this when we upgrade our app the Gatsby V2

### DIFF
--- a/app-web/gatsby-config.js
+++ b/app-web/gatsby-config.js
@@ -77,6 +77,7 @@ module.exports = {
       resolve: 'gatsby-transformer-remark',
       options: {
         plugins: [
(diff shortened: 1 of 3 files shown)


patricksimonian:
rebased ontop of master (had to force push)

patricksimonian:
forced push after rebase

PR #137, opened 2018-11-27
Refactor Card Components to render based on resource type

patricksimonian:
# About

This is the first of many refactors to the card components. Initially cards rendered with a big assumption that they were linked to 'github'. Now the cards render based on the resource type.

## Changes
- Created 4 new Card Components (All of them are identical at this point)
- `<Cards ../>` now renders Cards based on the resource type
- Created a `<Card />` higher order component to create some composition. This is also used as a predictable container for the visual layout of a Card


patricksimonian:
Any thoughts on the `<Card >` higher order component? I thought it may be a good way to abstract the common elements that all Card Components Share. At this point its only the style of the outer container but it could contain other common things

PR #138, opened 2018-11-27
Refactored Gatsby CreatePages to choose template based on source type

patricksimonian:
# About

The page creation initially was only based on mediaType (ie page templates for markdowns vs html).
This led to templates that were strongly tied to github sources. Since Siphon can assign different source types as they are supported. The way templates are chosen to create a page from a siphon node has been refactored.

## Changes
- Updated createPages logic to choose templates based on source type
- Renamed Template files to be more related for github source types

### DIFF
--- a/app-web/gatsby/createPages.js
+++ b/app-web/gatsby/createPages.js
@@ -20,13 +20,30 @@
 
 // create pages based on nodes
 const { resolve } = require('path');
+const { SOURCE_TYPES } = require('../plugins/gatsby-source-github-all/utils/constants');
 
-const getTemplate = key => {
+/**
+ * Get Templates based on source and media type
+ * @param {String} source 
+ * @param {String} mediaType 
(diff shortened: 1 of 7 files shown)


app-web/gatsby/createPages.js
```diff
@@ -20,13 +20,30 @@
...
+ * Get Templates based on source and media type
+ * @param {String} source 
+ * @param {String} mediaType 
+ * @returns {string} the path to the template
+ */
+const getTemplate = (source, mediaType) => {
   const TEMPLATES = {
-    'text/markdown': resolve(__dirname, '../src/templates/SourceMarkdown.js'),
-    'text/html': resolve(__dirname, '../src/templates/SourceHTML.js'),
+    [SOURCE_TYPES.GITHUB]: {
```
ShellyXueHan: I've got questions for you on this, lmk when to talk:)

PR #139, opened 2018-11-27
Bug: Adding ignore to frontmatter still creates a card that goes to no where

patricksimonian:
# About

We found an issue where markdown files may have an 'ignore' front matter property set to true, it would still create a card component which would link to a page that hadn't be created. 

This was found because there was some logic in the createPages to prevent pages from being created in the node but the card components graphql did not share a similar logic. To circumvent this issue all together, the approach i took was prevent graphql nodes from being created if they have an 'ignore: true' front matter property. 

## Changes
- metadata property now includes 'ignore'
- source nodes now filter out files that have a ignore: true metadata property
- added unit tests for said filtering function


patricksimonian:
force update on rebase to master

PR #145, opened 2018-11-28
Add images to cards

patricksimonian:
# About
added images to all card components

## Changes
- added images to cards
- restyled cards so that images fit predictably in containers when they exist
- added truncation of description text incase it's over 5 lines

### DIFF
--- a/app-web/src/components/Cards/Card/ComponentCard.js
+++ b/app-web/src/components/Cards/Card/ComponentCard.js
@@ -41,7 +41,7 @@ const ComponentCard = ({ sourcePath, sourceName, title, description, resourcePat
 ComponentCard.displayName = 'Component Card Component';
 
 ComponentCard.propTypes = {
-  sourcePath: PropTypes.string,
+  sourcePath: PropTypes.string.isRequired,
   sourceName: PropTypes.string.isRequired,
   title: PropTypes.string.isRequired,
   description: PropTypes.string.isRequired,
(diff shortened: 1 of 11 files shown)


patricksimonian:
not ready for merge until #143 is merged and **this pr is rebased ontop of master**

app-web/src/components/Cards/Card/ComponentCard.js
```diff
@@ -39,6 +45,11 @@ ComponentCard.propTypes = {
   title: PropTypes.string.isRequired,
   description: PropTypes.string.isRequired,
   resourcePath: PropTypes.string.isRequired,
+  image: PropTypes.string,
+};
+
+ComponentCard.defaultProps = {
+  image: '',
```
sutherlanda: Maybe set a default image path here? You could make the default image a configuration item like you did with maxTitleLines, maxDescription lines.

patricksimonian: yes! Thats a great idea, will be working on that on a new pull request. We are going to first identify what default images to use based on the card type :)

ShellyXueHan:
sourcePath for card components is required

app-web/src/components/Cards/Card/ComponentCard.js
patricksimonian: The react `<Image>` component will not render a 'broken image' if the image fails to load which is the fall back until the next PR

PR #152, opened 2018-12-01
fix an issue where twitter/opengraph urls that were web uris would not unfurl

patricksimonian:
This fixes the issue were twitter card/open graph images, that were formatted as valid web uris `(eg https://www..)`, would not unfurl correctly in siphon

### DIFF
--- a/app-web/plugins/gatsby-source-github-all/utils/plugins.js
+++ b/app-web/plugins/gatsby-source-github-all/utils/plugins.js
@@ -189,7 +189,6 @@ const externalLinkUnfurlPlugin = async (extension, file) => {
     try {
       metadata = await scrape(file.metadata.resourcePath);
     } catch (e) {
-      console.log(e);
       return file;
     }
     // metadata comes in with properties for each type of unfurl spec (twitter, openGraph etc)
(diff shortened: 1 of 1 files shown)


patricksimonian:
rebased onto master

app-web/plugins/gatsby-source-github-all/utils/plugins.js
```diff
@@ -189,20 +189,23 @@ const externalLinkUnfurlPlugin = async (extension, file) => {
...
     const combinedMetadata = { ...metadata.twitter, ...metadata.openGraph };
     // const metadata = await metascraper({ html, url });
     // update image to have resource path prepended to it if it is not https
     if (TypeCheck.isString(combinedMetadata.image)) {
-      if (validUrl.isWebUri(combinedMetadata.image) === undefined) {
-        combinedMetadata.image = url.resolve(file.metadata.resourcePath, combinedMetadata.image);
-      }
-
-      file.metadata.unfurl = createUnfurlObj(UNFURL_TYPES.EXTERNAL, combinedMetadata);
+      combinedMetadata.image = url.resolve(file.metadata.resourcePath, combinedMetadata.image);
```
jleach: `url.resolve` has caused me some grief in the past. If it does not get a URL it fails and in my case it was hard to track down where. It might be worth wrapping it in a try / catch unless you're pretty sure it won't get invalid params.

jleach:
@patricksimonian One comment that is food-for-though - no show stoppers. Approved.

app-web/plugins/gatsby-source-github-all/utils/plugins.js
patricksimonian: good point,  i didn't think of that. I'll add that try catch. Thanks for the wisdom there

patricksimonian: actually took a look, I think i should be in the clear. the top level if validates the resource path as a uri

PR #153, opened 2018-12-01
Feature: allow user to specify context directories in a GitHub repo to fetch files from

ShellyXueHan:


app-web/plugins/gatsby-source-github-all/utils/fetchSourceGithub.js
```diff
@@ -224,14 +224,33 @@ const applyBaseMetadata = (
   };
 };
 
+/**
+ * filters files by the context directories from source
+ * @param {Array} files the files
+ * @param {Object} contextDir array of paths to get files in a repo
```
patricksimonian: js doc should say type Array

app-web/plugins/gatsby-source-github-all/utils/fetchSourceGithub.js
```diff
@@ -224,14 +224,33 @@ const applyBaseMetadata = (
...
+ */
+const filterFilesByContext = (files, contextDir) => {
+  const targetPaths = contextDir.map(dir => {
```
patricksimonian: at one point we need to validate contextDir from the registry's the sourceproperties as it is an uncontrolled value

app-web/plugins/gatsby-source-github-all/utils/fetchSourceGithub.js
```diff
@@ -224,14 +224,33 @@ const applyBaseMetadata = (
...
+    return dir.charAt(0) === '/' ? dir.substring(1) : dir;
+  });
+  // now only return files in the target context dir array
+  const contextFiles = files.filter(file => {
+    return targetPaths.some(path => file.path.indexOf(path) >= 0);
```
patricksimonian: this will cause any paths to validate to true
if path is `docs` and file.path is `/foo/bar/docs` this will pass this test. 
We should explictily define context directories based on the root. if the example path wants to be set as a context dir
it should have a path in the registry set as `/foo/bar/docs`

patricksimonian: i'd suggest either one of two options
`indexOf(path) === 0` or create a regex object from file.path and test it 
```javascript
const re = new RegExp(`^${file.path}`);
return re.test(path);
```

app-web/plugins/gatsby-source-github-all/utils/fetchSourceGithub.js
patricksimonian: perhaps allows contextDir to have 1 of 2 states
- type check it as a string and then convert it into an array [string]
- type check that it is an array and do not continue filtering if it isn't an array

and also maybe we provide a default value for the contextDir in the event it doesn't exist?
(somewhere?)
```javascript
const filterFilesByContext = (files, contextDir = []) => { 
```

ShellyXueHan: right, implemented it to use option 1

app-web/plugins/gatsby-source-github-all/utils/fetchSourceGithub.js
ShellyXueHan: as the file paths are all relative to root, will get rid if extra `/` from contextDir and match with `indexOf(path) === 0`

docs/registry.sample.yml
```diff
@@ -19,4 +19,16 @@ repos:
       repo: 'devhub-signing-web'
       branch: my-branch
     resourceType: Self-Service Tools
+  - name: DevHub
+    sourceType: 'github'
+    sourceProperties:
+      url: https://github.com/bcgov/devhub-app-web/
+      owner: bcgov
+      repo: 'devhub-app-web'
+      context: ['/docs/test-context-dir', 'plugins/docs']
```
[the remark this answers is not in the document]
ShellyXueHan: thought this file wasn't being used, so kept it there as demo. Does this affect anything?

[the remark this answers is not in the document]
patricksimonian: ignoreeeee : P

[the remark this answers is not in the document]
patricksimonian: the only thing i'd suggest is yaml lists can be array looking but they also can look like
```yaml
context: 
  - dir1
  - dir2
  - dir3
```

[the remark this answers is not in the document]
ShellyXueHan: nice! updated

PR #155, opened 2018-12-03
Imp/add card actions

patricksimonian:
# About
First of few additions to the card layout

## Changes
- added the `<Ribbon />` and `<ActionsRibbon />` component


app-web/src/constants/ariaLabels.js
```diff
@@ -4,3 +4,6 @@ export const ARIA_LABEL_FILE_IN_REPO =
 export const ARIA_LABEL_RESOURCE = 'Learn more about this article';
 export const ARIA_LABEL_FILTER_SELECT = 'Select to filter cards by the resource type';
 export const ARIA_LABEL_TO_GITHUB_USER = 'Select to view this authors Github Profile';
+export const ARIA_LABEL_TO_GITHUB_ISSUES = "Select to view this respositories' issues";
```
ShellyXueHan: better to make descriptions more details by saying 'this Github repos....'

app-web/src/constants/ui.js
```diff
@@ -31,36 +69,16 @@ export const CARD_CONFIG = {
...
-    {
-      name: 'Watch',
-      flag: 'features.card.watch',
-      icon: faEye,
-      ariaLabel: ARIA_LABEL_WATCH_GITHUB_REPO,
-      getHref: getGithubWatchRoute,
-    },
-  ],
+  actionsRibbon: {
+    [RESOURCE_TYPES.RESPOSITORIES]: [
```
ShellyXueHan: why do you need the []?

app-web/src/constants/ariaLabels.js
patricksimonian: changed!

app-web/src/constants/ui.js
patricksimonian: Oh i'm setting the key to be exactly the same as the enum for resource types
that way if resource types are ever changed this config doesn't need to be changed

PR #156, opened 2018-12-03
Add plugin to automatically link repository type cards to github

patricksimonian:


app-web/plugins/gatsby-source-github-all/utils/constants.js
```diff
@@ -27,14 +27,22 @@ const SOURCE_TYPES = {
...
+};
 // Resource Types for Devhub
-const RESOURCE_TYPES = [
-  'Documentation',
-  'Repositories',
-  'People',
-  'Projects',
-  'Self-Service Tools',
-  'Components',
+const RESOURCE_TYPES_LIST = [
```
jleach: You can create this as needed with `Object.values(RESOURCE_TYPES)` it has the small advantage of not having to maintain two lists.

app-web/plugins/gatsby-source-github-all/utils/constants.js
```diff
@@ -162,6 +170,7 @@ module.exports = {
   MEDIATYPES,
   SOURCE_TYPES,
   RESOURCE_TYPES,
+  RESOURCE_TYPES_LIST,
```
jleach: Object.values(RESOURCE_TYPES),

app-web/plugins/gatsby-source-github-all/utils/plugins.js
```diff
@@ -234,11 +234,25 @@ const markdownPersonaPlugin = async (extension, file, { personas }) => {
   return file;
 };
 
+/**
+ * sets resource path to be path to file in repo if resource type is repo
+ * @param {String} extension 
+ * @param {Object} file 
+ * @returns {Object} the modified file
+ */
+const repositoryResourcePathPlugin = async (extension, file) => {
```
jleach: Does this one need to be `async`?

patricksimonian: it does. Unfortunately, the only way to allow true async functions to coexist with sync functions, is to have all of them return a promise.

app-web/plugins/gatsby-source-github-all/utils/constants.js
patricksimonian: great point. Changed.

app-web/plugins/gatsby-source-github-all/utils/plugins.js
jleach: Is that because of the way it's used `.use(repositoryResourcePathPlugin)`?

patricksimonian: this is in respect to the usage of this mini framework. It is a serial pipeline utility. Plugins make a change and feed to the next plugin.

patricksimonian: kind of. It more to control running these plugins serially. The .use is just a friendly api i felt. Any thoughts on it?

patricksimonian: The plugins running in order **may sometimes** be beneficial. Indeed, there are many plugins that have an assumption of order (kind of like a dependancy on some plugins having already been applied). For example, we validate front matter properties with the first plugin in the `fetchFileFromRepo fn`.

jleach: I think what you're telling me is all the functions in `.use()` need to return a promise for this to work. Otherwise I still don't see why its needed; its not doing anything with callbacks internally so it will run synchronously. Await just pause execution without blocking the runloop to allow async code to complete.

patricksimonian: Yes i see your point. Here is the actual api for the transformer

```javascript
const fileTransformer = (fileExtension, originalFile) => {
  let file = clone(originalFile);
  const plugins = [];
  return {
    use(plugin, options = {}) {
      // using typeof check instaed of TypeCheck since async functions are nott
      // true instances of Function
      if (typeof plugin !== 'function') {
        throw new Error('Plugin must be function');
      }
      plugins.push([plugin, options]);
      return this;
    },
    async resolve() {
      for (const plugin of plugins) {
        file = await plugin[0](fileExtension, file, plugin[1]);
        if (file === undefined) {
          throw new Error('Plugin must return the file');
        }
      }
      return file;
    },
  };
};
```
having the `.use()` return a promise would allow the plugins to seperately not *have* to return a promise but then the usage of the transformer pipeline would have to be refactored. Right now, `.use()` returns `this` so that it can be chained. Do you think I should change it over? it's a simple enough change i think, the `.use` chaining could completly be replaced by perhaps an array of plugins in the file transformer function eg `await fileTransformer(fileExtension, file, plugins = [])`

patricksimonian: a user story has been created to resolve this technical debt

PR #157, opened 2018-12-03
add metadata ribbon

patricksimonian:
for now this only contains resource type. I do know shea has a plan to add the 'persona' attributes. It should be easy enough to integrate personas into this ribbon


ShellyXueHan:
update unit tests

patricksimonian:
@ShellyXueHan added unit tests + snaps

PR #165, opened 2018-12-05
Imp/filter nav

patricksimonian:
## Summary
The Select List for filtering was replaced with the Design System's navigation component


patricksimonian:
closing to fix history

PR #166, opened 2018-12-05
updated the table headers so they don't break a word before its compl…

rajpersram:
## Summary
Added one line into the class for header th to prevent words from wrapping before the word's completion.

## Notable Changes 
one line update on table th


patricksimonian:
rebased onto master

PR #173, opened 2018-12-10
Feature: config siphon to take specific source registry yaml file

ShellyXueHan:
## Summary
Siphon is originally taking the source from registry.yml file.
we want to be able to config this, so that different registry.yml files would be picked in different environments.
This would allow the deployment in TEST environment to use fixtures and make test cases more stable.


patricksimonian:
is this not ready for merge because it needs to be rebased?

patricksimonian:
definitely needs to be rebased now.

ShellyXueHan:
@patricksimonian did a rebase, will use `--interactive` next time to squash the commits.

app-web/gatsby-config.js
```diff
@@ -70,6 +71,7 @@ module.exports = {
       options: {
         tokens: {
           GITHUB_API_TOKEN: process.env.GITHUB_TOKEN,
+          SOURCE_REGISTRY_TYPE: process.env.REGISTRY_TYPE || 'SourceRegistryYaml',
```
patricksimonian: I think this should be moved out of tokens and be stand alone

patricksimonian: also could we should perhaps document this somewhere just so we know how the source registry is acutally sourced. Its not too important now, but if this is ever exported as an npm package we will need explain the dependancy on gatsby source filesystem and gatsby transformer yaml plugins

PR #177, opened 2018-12-11
Update index page to organize nodes by collection name

patricksimonian:
## Summary
Updated the graphql query. Also fixed a minor issue in some docs.

### DIFF
--- a/app-web/src/pages/index.js
+++ b/app-web/src/pages/index.js
@@ -30,7 +30,7 @@ export class Index extends Component {
         .map(node => ({
           ...node.unfurl,
           resourcePath: node.resource.path,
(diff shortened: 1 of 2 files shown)


app-web/src/pages/index.js
```diff
@@ -135,6 +135,10 @@ export const resourceQuery = graphql`
           attributes {
             persona
           }
+          collection {
```
jleach: GitHub is showing these in dark-read. This usually means a syntax error but I don't see any in the string.

patricksimonian: yeah i'm not sure why this is happening.. ill take a look into it

PR #186, opened 2018-12-13
Update deploy config to take #replicas by configuration

patricksimonian:
Deploy config had replicas hardcoded to 2. 

Because of our github flow, all of the dev deployments would run with 2 replicas by default. In the event the dev deployment isn't cleaned up (after successful merge of a pr), having 2 replicas eats up ram faster than 1 would. This causes issues later down the road when attempting to deploy/redeploy later pull requests because the dev namespace has exceeded its quota for ram.


patricksimonian:
closing as this seems to not be the way to solve this issue

PR #191, opened 2018-12-14
Add configurations for docker, so we can develop in a consistent local env

patricksimonian:
## Summary
Add docker file and updated readme with instructions on how to use

## Notable Changes 
- **package.json npm run dev script** has the `-H 0.0.0.0` paramater appended to it. This will allow for docker to run locally. The caveat is now local development on a windows machine is not possible. So sad.

### DIFF
--- a/app-web/Dockerfile
+++ b/app-web/Dockerfile
@@ -11,5 +11,5 @@ RUN npm ci --silent
 
 COPY . .
 
-EXPOSE 8000 8000
+EXPOSE 8000
 CMD [ "npm", "run", "dev" ]
--- a/app-web/Dockerfile
+++ b/app-web/Dockerfile
@@ -7,9 +7,9 @@ RUN mkdir /usr/src/app
 WORKDIR /usr/src/app
 
 COPY package*.json /usr/src/app/
-RUN npm ci --silent
-
 COPY . .
+
+RUN npm ci --silent
 
 EXPOSE 8000
 CMD [ "npm", "run", "dev" ]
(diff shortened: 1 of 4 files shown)


app-web/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM node:8
```
jleach: Here is some stuff I often add to my Dockerfiles:
```
ARG NODE_MAJOR_VERSION=10
ARG NODE_VERSION=10.14.x
ARG APP_VERSION="0.1.0"

ENV INSTALL_PATH /opt/app
ENV SUMMARY="MDMS Device Provisioning API ${APP_VERSION}"  \
  DESCRIPTION="MDMS Device Provisioning API {APP_VERSION} running node nodejs ${NODE_VERSION} on Alpine Linux"
ENV LC_ALL=en_US.UTF-8 \
  LANG=en_US.UTF-8 

LABEL summary="$SUMMARY" \
  description="$DESCRIPTION"
```

Then I'd use $INSTALL_PATH as a variable in line like `WORKDIR $INSTALL_PATH`

app-web/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+RUN apt-get update && \
+  yes | apt-get upgrade
+
+RUN mkdir /usr/src/app
+WORKDIR /usr/src/app
+
+COPY package*.json /usr/src/app/
```
jleach: I think this line isn't needed any more because `COPY . .` will take everything with it.

app-web/Dockerfile
patricksimonian: added those in, thanks!

app-web/Dockerfile
patricksimonian: removed it

PR #193, opened 2018-12-17
Imp/multi filter

patricksimonian:
## Summary
Feature to add multi filter selecting.

The filtering works in two levels. 
Primary filtering is A-Track style and filters over resource types

Secondary Filtering is more checkbox like
-
/bot-ignore-length


app-web/src/store/reducers/siphon.js
```diff
@@ -17,32 +17,108 @@ Created by Patrick Simonian
...
   error: false,
   messages: [],
+  filters: [],
+  filterGroups: defaultFilterGroups,
 };
 
+/**
+ * filters through the already filtered nodes with additional
+ * filters found from the filters list
+ * @param {Array} filteredNodes
```
ShellyXueHan: update para names: secondaryFilteredNodes & filters

patricksimonian: fixed!

PR #197, opened 2018-12-18
Refactor redux logic for filtering resources & updated tests

patricksimonian:
## Summary
After taking a deep look into how the secondary filtering would have worked, I refactored the reducer and greatly simplified it. This has some big performance benefits as implementing the filtering into the ui before these changes would cause many re renders of the ui for no reason.

## Notable Changes 
- refactored all redux logic for filtering secondary filtered nodes
- added a new action to reset all secondary filters
- I noticed that in the `siphonReducer.test.js` there was high dependancy between each test. Things like setting up values for a test would involve calling reducer actions outside of the unit that was being tested. **this has been resolved**


app-web/src/store/reducers/siphon.js
```diff
@@ -27,34 +23,57 @@ const initialState = {
...
-  filters: [],
-  filterGroups: defaultFilterGroups,
+  filters: defaultFilterGroups,
 };
 
 /**
  * filters through the primary filtered nodes by filters
  * found from the state.filters list
+ * checks if the dot prop matches the value for a given siphon node
+ * @param {Object} node
```
jleach: These will be hard to fill in later.

app-web/src/store/reducers/siphon.js
```diff
@@ -63,18 +82,59 @@ const findFilterGroup = (state, key) => {
  * sets 'active' property on a filter config object
  * @param {Object} state
  * @param {String} key
+ * @param {Boolean} isActive
```
jleach: These will be hard to fill in later.

app-web/src/store/reducers/siphon.js
```diff
@@ -63,18 +82,59 @@ const findFilterGroup = (state, key) => {
...
-const toggleFilter = (state, key) => {
+const toggleFilter = (state, key, isActive) => {
   const newState = { ...state };
-  const fg = findFilterGroup(state, key);
+  const fg = findFilter(state, key);
+  fg.active = isActive;
 
-  fg.active = !fg.active;
-  const newFilterGroups = newState.filterGroups.map(f => {
+  const newFilters = newState.filters.map(f => {
```
jleach: could probably be `map(f => f.key === fg.key ? fg : f);`

app-web/src/store/reducers/siphon.js
patricksimonian: good point fixed

app-web/src/store/reducers/siphon.js
patricksimonian: Definitely. Thanks!

PR #265, opened 2019-01-07
Siphon Collections as a first class object ( pr 1 of n)

### LINKED ISSUE #260: Refactoring so "collection" is a first-class object (rather than simply a grouping function as it is now) and attributes can be attached to it.
This is required to support features such as collection description, collection-specific template, and collection overview.

patricksimonian:
## Summary
Update Siphon to create collection nodes that reference their child siphon nodes.
In this pr, the fetch queue fn was refactored to create a list of 'collections'. This list is used looped over and sources fetched in a similar fashion leading to a backwards compatible way of fetching sources. 

The over all work required to convert siphon collections to a first class object is not entirely extensive but is best broken into smaller PRs. Therefor, this is the first of n successive prs.

Issue: #260


patricksimonian:
rebased onto master and squashed commits then force pushed

PR #286, opened 2019-01-14
Fix card show more/less toggle button to render an appropriate number of cards in groups of 3

### LINKED ISSUE #244: Should show full "rows" of cards in typical desktop view (currently first row show 3 cards, second is a single card)
Previously we had 4 cards across, and no second row. We should revert to that layout or something equally sensible.

patricksimonian:
## Summary
The number of cards that rendered within a collection in the devhub would sometimes show up in a less than ideal fashion. There maybe singled out cards that would wrap to a new row. This PR forces all cards to show up in atleast a group of three (if numCard >= 3);

## Notable Changes 
- added a helper to calculate the ideal card limit for toggling
- added unit tests for said helper

Fixes #244

### DIFF
--- a/app-web/src/components/Cards/Cards.js
+++ b/app-web/src/components/Cards/Cards.js
@@ -8,7 +8,7 @@ import RepositoryCard from './Card/RepositoryCard';
 import SelfServiceCard from './Card/SelfServiceCard';
 import ComponentCard from './Card/ComponentCard';
 import Toggle from './Toggle';
-import { CARD_TOGGLE_LIMIT, RESOURCE_TYPES } from '../../constants/ui';
+import { CARD_TOGGLE_LIMIT, RESOURCE_TYPES, CARDS_PER_ROW } from '../../constants/ui';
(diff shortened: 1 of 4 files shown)


patricksimonian:
@lukegonis you may be interested in the logic that was used to achieve how the cards render. It's all math! Gota love it 😸

app-web/src/components/Cards/Cards.js
```diff
@@ -92,17 +106,27 @@ const Cards = ({ topic, sourcePath, cards }) => {
       </Link>
     );
   if (cardComponents.length > 0) {
+    // find the ideal large screen limit based on number of cards, and cards rendering per row
+    const CARDS_PER_ROW = 3; // ideally in a true grid system we would have a more concrete styling
```
jleach: I might try constants like this in a `constants.js` file.

patricksimonian: good point, changed

PR #289, opened 2019-01-15
Change  npm install to ci in s2i script

patricksimonian:
## Summary
changed `npm install` to `npm ci` in assemble script to speed up build times.


patricksimonian:
wow yes. I think it doesn't i'll double check

patricksimonian:
yep. confirmed the current s2i base image is at npm version `5.6.0` and we need `5.7.0` to run npm ci.

patricksimonian:
@cvarjao can you point me to where i could get more information on making an s2i image with a higher node and npm version? not that I am looking to make one right now but it would be good information to know and start a discussion here.

PR #290, opened 2019-01-15
implement siphon node sorting by position

### LINKED ISSUE #257: Ability to maintain order for cards/nodes in card list view and side navigation (and elsewhere).

patricksimonian:
## Summary .
#287 
This it the front end implementation of the ordering.

Note that the graphql query auto orders collections by the sort function

### why don't you order the nodes within the collections in the graphql query?
tried it. didn't work. There doesn't seem to be a sort function allowed on chiild gatsby nodes. The ordering is done withini the siphon reducer loadCollections action.

## Notable Changes 
- used the new _metadata.position property of siphon nodes to create a order number based on the weight of the positions
```js
// eg position is an array of [collection {Number}, source {Number}, resource {Number}]
// weighting is based on the position of each element in the array. Left most are weighted heigher than 
// right most 
const position1 = [0, 2, 2];
const position2 = [1, 2, 2];
// after processing...
position1AfterWeight === '100.20.2'
position2AfterWeight === '200.30.3'
// not that the position is an 'address' this allows for lexographic sorting
```
Fixes #257

### DIFF
--- a/app-web/src/store/reducers/siphon.js
+++ b/app-web/src/store/reducers/siphon.js
@@ -36,20 +36,18 @@ const mapWithCallback = (array, cb) => array.map(cb);
  * weighting decreases left to right
  * @param {Array} position
  * @returns {String} an address of the position based on weight
- * [0, 2, 5] = 100.30.6
+ * [0, 2, 5] => '100.30.6' (this allows for easy lexographic sorting)
  */
-export const getTruePositionFromWeightedScale = position => {
-  return position
+export const getTruePositionFromWeightedScale = position =>
+  position
     .reduce((pos, val, index) => {
       const power = position.length - index - 1;
-
       const multiplier = val + 1; // since positions start at 0, we are adding 1 so that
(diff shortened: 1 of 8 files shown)


app-web/src/store/reducers/siphon.js
```diff
@@ -29,6 +29,28 @@ const initialState = {
...
+/**
+ * from an array of siphon positions
+ * [collection, source, resource]
+ * we return an integer by weighting each index to a power of ten
+ * weighting decreases left to right
+ * @param {Array} position
+ * @returns {String} an address of the position based on weight
+ * [0, 2, 5] = 100.30.6
+ */
+export const getTruePositionFromWeightedScale = position => {
```
jleach: Often my linter tells me to not put `{ }` around a return statement but rater delete them and the word `return`.

patricksimonian: great point, thanks. I've changed that.

PR #291, opened 2019-01-16
registry changes to set up a collection for service definitions

sheaphillips:
## Summary

Adding a new collection for Service Definitions, and the first of these for the gov OpenShift Container Service.

## Notable Changes 

- mod to registry to create new collection and explicitly include the OpenShift service def and explicitly ignore service definition from the existing openshift-wiki collection
- related change is to openshift-wiki repo to add the md file containing the service definition content.


patricksimonian:
i thought there may be a conflict with the lock file as it appears to have changed in the pr. perhaps deleting and running a reinstall after your rebase will fix it

PR #295, opened 2019-01-16
Add an 'overview' template as proof of concept for Collection Templating

### LINKED ISSUE #250: Custom / different templates for generated pages (initially for overview content)
We should dig into having multiple templates for pages generated in the framework.  The first use case would be for collections "overviews" where we would prefer something other than the current "sidebar" nav as a way to move between / visualize pages in the collection.

patricksimonian:
## Summary
Added a new template to devhub, the github 101 collection now renders ***slightly*** differently compared to others. This involved renaming the template files so that they follow a type of convention more or less. 

Fixes #250


patricksimonian:
not ready to merge until #294 is merged and this rebased ontop of it

PR #314, opened 2019-01-22
Primary filtering switched to true navigation

patricksimonian:
## Summary
Converted 'all resource type like pages' into actual pages. 

There is tons of duplicate could which is an evil of static sites. Some thought was put into refactoring the duplicated components into one **large** higher order component but i've decided to leave that out now as having resource types as pages allows for some flexibility in rendering possabilities

#313 #307 

/bot-ignore-length


patricksimonian:
not ready for merge until #313 has been merged and this rebased onto it.

PR #330, opened 2019-01-28
add Search component based on spec

patricksimonian:
## Summary
UI search bar component, created, tested but not implemented.

#307 

/bot-ignore-length


app-web/src/components/Search/Search.js
```diff
@@ -0,0 +1,71 @@
...
+import PropTypes from 'prop-types';
+import Button from '../UI/Button/Button';
+
+export class Search extends Component {
+  state = {
+    touched: false,
+  };
+
+  shouldComponentUpdate(nextState, nextProps) {
+    return nextProps.terms !== this.props.terms;
```
ShellyXueHan: isn't `terms` a string? So if not empty, set prop to empty?

PR #332, opened 2019-01-29
Implement reducer functions for search

patricksimonian:
## Summary
Added the search reducer as well as a big refactor to the reducer and tests so that they are more performant and stable

#307 
/bot-ignore-length


patricksimonian:
i swear its all snapshot

PR #345, opened 2019-01-30
Add a react link component that sets active class based on search params

patricksimonian:
## Summary
This is a component that wraps the already in use `<Link>` component. Essentially it overrides the links behaviour of setting active class based on path matching. Instead it sets 'active' based on if query params match

__please note__ at this time the `<SearchAwareLink />` will only set active if it has one search parameter in its 'to' prop. multiple search param matching hasn't been implemented

## So why do we need this?

This component is a refactor with the logic used in #344 which allows links to be active based on what searchs are currently in use. The primary navigiation filters are essentially a set of canned searches. #344 addressed a bug where the active link styling was not being applied because it wasn't aware of query params.

### DIFF
--- a/app-web/src/components/UI/Link/SearchAwareLink.js
+++ b/app-web/src/components/UI/Link/SearchAwareLink.js
@@ -19,13 +19,14 @@ import queryString from 'query-string';
 import React from 'react';
 import PropTypes from 'prop-types';
 import Link from './Link';
-import Search from '../../Search/Search';
 
-// bootstraps the gatsby link, when the 'to' prop is a relative path,
-// the active class is set based on if the if the query string parameter matches
-// this is not a replacement for the Link component, as setting activeClassName by path matching
-// doesn't work anymore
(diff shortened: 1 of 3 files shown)


app-web/src/components/UI/Link/SearchAwareLink.js
```diff
@@ -0,0 +1,66 @@
...
+ * returns true if the to attributes search param exists within the current location search params
+ * @param {Object} location the location prop for the current windo
+ * @param {String} to the actual href used by the link
+ * @returns {Boolean}
+ */
+export const linkMatchesQueryString = (location, to) => {
+  // convert to into an URL object
+  // at this point we absolutely know that 'to' must be a relative path,
+  // there for we prepend a dummy url to it so that it can be processed by URL
+  const url = new URL(`https://developer.gov.bc.ca/${to}`);
```
jleach: URLs might be better in a const file so they can be changed and reused easily.

PR #348, opened 2019-01-31
Fix issue where search bar allowed for broken searches

patricksimonian:
## Summary
If you entered a query in the search bar such as `foo:bar`, lunr would actually take that a 'presence token' and throw because there is no index field `foo` within the index.

This was found to be caused by the search bar passing in the query into the literal search function. 

A few changes were made so that the search bar query is taken literally as a string and isn't parametrized, in addition a extra query param was passed along with the resource type links so that they are parametrized on search


patricksimonian:
not ready for merge until integrated with #344

PR #360, opened 2019-02-01
Adjust max width of markdown page content container

patricksimonian:
##Summary 

Set max width of markdown container to 800px for better readability.


patricksimonian:
rebased for merge

PR #376, opened 2019-02-06
Improve tech debt for search feature

### LINKED ISSUE #349: Improve Unit/Integration testing for search feature on homepage
No integration testing has been done for the search feature on the homepage. We should add some as this feature is more then likely going to evolve over time. We should make sure base behaviour remains consistent.

patricksimonian:
## Summary
Fixes #349

### DIFF
--- a/app-web/src/pages/index.js
+++ b/app-web/src/pages/index.js
@@ -204,7 +204,7 @@ const mapDispatchToProps = dispatch => {
     setSearchResults: results => dispatch(actions.setSearchResults(results)),
     hideWelcomeMessage: () => dispatch(actions.setWelcomePanelViewed(true)),
     setSearchQuery: query => dispatch(actions.setSearchQuery(query)),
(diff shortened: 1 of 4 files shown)


app-web/src/pages/index.js
```diff
@@ -204,7 +204,7 @@ const mapDispatchToProps = dispatch => {
     setSearchResults: results => dispatch(actions.setSearchResults(results)),
     hideWelcomeMessage: () => dispatch(actions.setWelcomePanelViewed(true)),
     setSearchQuery: query => dispatch(actions.setSearchQuery(query)),
-    setSearchBarTerms: resourceType => dispatch(actions.setSearchBarTerms(resourceType)),
+    // setSearchBarTerms: resourceType => dispatch(actions.setSearchBarTerms(resourceType)),
   };
```
ShellyXueHan: do you still need this in the future? maybe comments?

patricksimonian: uncommented thanks for checking that out@

PR #377, opened 2019-02-08
Added Noto Sans font file

dlevineBC:
## Summary
We need to use a locally hosted version of the google font. I've included this into the assets folder with a new folder called fonts.

## Notable Changes 
- 
-
-


dlevineBC:
@patricksimonian I've removed the google api link from the html.js file. That was the only instance I've found of calling the font from a link

patricksimonian:
> @patricksimonian I've removed the google api link from the html.js file. That was the only instance I've found of calling the font from a link

awesome !

PR #378, opened 2019-02-08
Imp/design system order

patricksimonian:
## Summary
Change registry file for design system to declare components in order. This exposed a bug where ordering would break if the # resources > 10. This bug was rectified in the pr as well.

### DIFF
--- a/app-web/plugins/gatsby-source-github-all/utils/helpers.js
+++ b/app-web/plugins/gatsby-source-github-all/utils/helpers.js
@@ -369,9 +369,23 @@ const assignPositionToResource = source => (resource, position) => ({
   },
 });
 
+/**
+ * creates a string that is usable in a natural sort
+ * it padds numbers with zeros so that it will sort normally
+ * ie [0, 0, 1], [0, 0, 10] => 0000.0000.0001 , 0000.0000.0010
+ * @param {Array} pos the position array
+ */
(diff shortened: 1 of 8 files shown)


patricksimonian:
@dlevineBC if you take a look at the registry file (`registry.yml`) in the files changed tab. You'll see how i modified the registry to explicitly declare the files in the order you wanted

app-web/plugins/gatsby-source-github-all/utils/helpers.js
```diff
@@ -369,9 +369,23 @@ const assignPositionToResource = source => (resource, position) => ({
   },
 });
 
+/**
+ * creates a string that is usable in a natural sort
+ * it padds numbers with zeros so that it will sort normally
+ * ie [0, 0, 1], [0, 0, 10] => 0000.0000.0001 , 0000.0000.0010
+ * @param {Array} pos the position array
+ */
+const convertPositionToSortableString = (maxPadding, pos) =>
```
ShellyXueHan: maybe consider hex value? aim at bigger than 9999 ;P

patricksimonian: Thats a really good point. I won't change it for now though. I have a feeling we will be changing how positions are maintained with the upcoming graph structure. which would make this function useless at that point:P

PR #383, opened 2019-02-09
bugs/380-381-componentpreview-bugs

### LINKED ISSUE #380: ComponentPreview component does not resize based on size of content.
Content that is "larger" than the default/fixed size of the iframe the ComponentPreview component creates will cause horizontal and vertical scrollbars to be shown, in some cases masking the actual intended content (e.g. the text area preview).

### LINKED ISSUE #381: ComponentPreview breaks when source uses default/master branch.
The ComponentPreview component does not handle the case where an explicit branch is not specified in the registry source def (so shows up as "null" in the graph).  We need to provide 'master' as the default to handle this case.

The behaviour is that an error message from the component shows up instead of the expected rendering of the component.

sheaphillips:
- Modifications so that `master` is used as a default branch  to pull ifrmae content from when none is provided or can be extracted from the source.
- Add support for seeting arbitrary properties on on ComponentPreview and having tham passed through to the iframe.  `height` and `width` are the most obvious ones to help with bcgov/design-system#104

closes #381
close #380

### DIFF
--- a/app-web/src/components/ComponentPreview/ComponentPreview.js
+++ b/app-web/src/components/ComponentPreview/ComponentPreview.js
@@ -28,54 +28,55 @@ export default class ComponentPreview extends React.Component {
     };
   }
 
-  async componentDidMount() {
+    async componentDidMount() {
 
-    //pull in the values to retrieve to preview content from GitHub from props, if provided
-    let { owner, repo, path, branch } = this.props;
+        //pull in the values to retrieve to preview content from GitHub from props, if provided
+        let {owner, repo, path, branch} = this.props;
 
-    //allow for default values - relative to file that has embedded the component -  if values for props not provided
-    owner =  owner || this.props.node.source._properties.owner;
-    repo = repo || this.props.node.source._properties.repo;
-    branch = branch || this.props.node.source._properties.branch || 'master';
+        //allow for default values - relative to file that has embedded the component -  if values for props not provided
+        owner = owner || this.props.node.source._properties.owner;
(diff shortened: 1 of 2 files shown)


app-web/src/components/ComponentPreview/ComponentPreview.js
```diff
@@ -28,54 +28,55 @@ export default class ComponentPreview extends React.Component {
...
-    let { owner, repo, path, branch } = this.props;
+        //pull in the values to retrieve to preview content from GitHub from props, if provided
+        let {owner, repo, path, branch} = this.props;
 
-    //allow for default values - relative to file that has embedded the component -  if values for props not provided
-    owner =  owner || this.props.node.source._properties.owner;
-    repo = repo || this.props.node.source._properties.repo;
-    branch = branch || this.props.node.source._properties.branch;
+        //allow for default values - relative to file that has embedded the component -  if values for props not provided
+        owner = owner || this.props.node.source._properties.owner;
```
patricksimonian: Can we document what properties are passable/are expected to this component?
The React PropTypes would be a great use case for that here.

sheaphillips: thanks for the suggestion. I've added PropTypes to the component to provide validation and some minimal doco.  Room for improvement on the doco side for sure.

patricksimonian:
@sheaphillips ready to promote this to prod?

sheaphillips:
yes. clicking now.

PR #384, opened 2019-02-11
Update registry.yml

dlevineBC:
## Summary
There was one page that was not in the correct order, so I updated the registry file

## Notable Changes 
-
-
-


patricksimonian:
@sheaphillips Personally I'd be okay merging this directly without going through the pipeline since to avoid all of this rebasing business. Any thoughts?

PR #419, opened 2019-02-21
Imp/docker compose

patricksimonian:
## Summary
Added Docker Compose file for local development. We were having issues with no Hot module reloading when running the dev server. Using docker compose fixed this and provides a much easier developer experience when trying to get set up in minimal time.


patricksimonian:
#404

PR #427, opened 2019-02-22
Card Layout and Designs Updated

patricksimonian:
## Summary
Redesigned Cards Yay!

## Notable Changes 
-  Integrated CSS-in-JS
-  changed cards to be in a bootstrap grid layout

/bot-ignore-length


app-web/src/components/Cards/Toggle.js
```diff
@@ -37,17 +39,27 @@ class Toggle extends Component {
       </Button>
     );
     const { cardComponents, cardLimits } = this.props;
-    const cardShow = this.state.toggled ? cardComponents.length : cardLimits;
+    let maxRows = Math.ceil(cardComponents.length / cardLimits);
+
+    let rows = [];
```
patricksimonian: this 'row' thing will be canned in an upcoming PR as we move to the homepage not showing all resources but only a predefined set of resources (4) and collections(2)

patricksimonian: this is why I don't have anyhting comments, this was a change to 'get it working for now'

PR #446, opened 2019-03-04
Refactor Search Function and combine partial/full searching to get better search results

patricksimonian:
## Summary
as pointed out by @jleach searching 'mobile' or 'signing' did not retrieve the mobile signing card as you would expect. This was found to be because searches were being conducted by partial tokenization eg __'*sign*'__.

I modified the search query to search in two ways, using partial and full matches and then combined the results to cast a larger net for relevant searches.


patricksimonian:
rebased and ready to merge

PR #450, opened 2019-03-05
Add Collection Preview Component

patricksimonian:
## Summary
Created the Component for the collection preview to be used in the collection view page

### DIFF
--- a/app-web/src/store/selectors/collection.js
+++ b/app-web/src/store/selectors/collection.js
@@ -31,7 +31,11 @@ export const selectCollectionsWithResources = createSelector(
   (collections, resources) =>
     collections.map(collection => ({
       ...collection,
-      resources: collection.resources.map(id => ({ ...resources.resources.byId[id] })),
+      resources: collection.resources.map(resource => {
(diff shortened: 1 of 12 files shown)


app-web/src/store/selectors/collection.js
```diff
@@ -31,7 +31,11 @@ export const selectCollectionsWithResources = createSelector(
   (collections, resources) =>
     collections.map(collection => ({
       ...collection,
-      resources: collection.resources.map(id => ({ ...resources.resources.byId[id] })),
+      resources: collection.resources.map(resource => {
+        // collections are initialized with resources only having { id: '...' }
+        // this maps the entire resource object to the collection
+        return { ...resources.resources.byId[resource.id] };
```
ShellyXueHan: why is it `resources.resources.xxx`?

patricksimonian: hmm could be better naming but the global redux state looks liked 
```
{
  auth: ..
  resources: ..
  ui: ..
}
```
the resources reducer has a property called resources
```
{
resourcesLoaded: ..
availableResources:..
resources: ..
}
```

PR #481, opened 2019-03-20
Refactor search portions of the resources reducer

patricksimonian:
## Summary
Removed the setSearchBar terms, it was redundant. The 
`searchbarTerms` and `query` states in that reducer have a similar behaviour and can be combined. 

## Notable Changes 
- removed duplicate state
- addressed some prop type errors that have been plaguing dev


app-web/src/components/Cards/Card/Card.js
```diff
@@ -36,7 +36,7 @@ import { RESOURCE_TYPES_LIST } from '../../../constants/ui';
 
 const Card = ({ type, title, description, image, link, theme }) => {
   // console.log(theme, type);
-  let isExternal = validUrl.isWebUri(link);
+  let isExternal = !!validUrl.isWebUri(link);
```
patricksimonian: this was to address a proptype error, this function is turns out does not return a boolean when a link __is__ a web uri, it returns a string instead. using `!!` is an easy way to cast falsy/truthy values

PR #490, opened 2019-03-21
Search Feedback and Resetting

patricksimonian:
## Summary
When Searching there was no feedback for the user explaining what had been searched and how to clear it. A simple  'search pill' has been integrated to reset the search as well as give visual feedback that a search has occurred. 

/bot-ignore-length

Fixes #465

PR #491, opened 2019-03-21
Search/reset query building

patricksimonian:
## Summary
Add utilities for tokenizing search queries into groupings of logical AND term presence as well as logic OR term presence.

The idea is that a search will be broken up into into two groups the original search with all terms maintaining a term presence predicate IE

Searching for '__API docs for Node js__' will translate to `'API+ docs+ for+ Node+ js+'` 
this means that all terms must be present in the document to succeed

As well as a grouping of tokenized terms

The terms will then also be broken up into a tokenized form with all stop words, punctation filtered

Searching for '__API docs for Node js__' will translate to `['API', 'docs', 'Node', 'js']` 

These two groupings are combined to produce a search pills
[entirequery, token1, token2, token3, ...etc]

## Notable Changes 
- added integration tests for tokenization of search
-

### DIFF
--- a/app-web/__tests__/utils/search.test.js
+++ b/app-web/__tests__/utils/search.test.js
@@ -1,4 +1,4 @@
-import { getSearchResults } from '../../src/utils/search';
+import { getSearchResults, tokenizer, getGroupedQueriesForLunr } from '../../src/utils/search';
 
 describe('Search Helpers', () => {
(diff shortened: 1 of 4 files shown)


patricksimonian:
not ready for merge until #490 is merged and this rebased ontop of it

patricksimonian:
closing as this pr has now gone stale.

PR #535, opened 2019-04-02
Update registry.yml

orinevares:
## Summary
Adding in some new component documentation for the design system


patricksimonian:
can you confirm your content is found and looking the way you expect it too ?
### Dev URL below
https://devhub-static-dev-535-devhub-dev.pathfinder.gov.bc.ca/

orinevares:
Looks good! I'll just need to adjust the live code snippet iFrame in the MD file since its too small on a couple pages right now.

PR #545, opened 2019-04-05
Imp/520 integrate eventbrite

### LINKED ISSUE #561: Prepare Eventbrite PR For Release
> This is typically not the case 'that we need to prepare a feature for release'

The eventbrite #545 feature has brought up a few things that we want to address prior to releasing.

- How are events created and their data annoted within the platform #558? We need to understand this so we do not inadvertantly display private events 
- How authentication works within the eventbrite api. Ideally there should be a personal access token that is linked to the CSI Lab event brite app. 
- Provide instructions on how to obtain said token for any user who is downloading the application
- Provide provisions to not need the eventbrite access token (remove the event feature if the token does not exist) 
- update ui to let user know they are going to Eventbrite

patricksimonian:
## Summary
Initial PR for the 'Events' page MVP

## To Do

- [x] -convert event page into template, if no access token is provided events page is not created

Fixes #561


app-web/src/components/Event/index.js
```diff
@@ -0,0 +1,126 @@
...
+      </CardWrapper>
+    </LinkWrapper>
+  );
+};
+
+Event.propTypes = {
+  title: PropTypes.string.isRequired,
+  where: PropTypes.string,
+  organizer: PropTypes.string.isRequired,
+  logo: PropTypes.string,
```
ShellyXueHan: default value?

PR #602, opened 2019-05-06
Add sizzy card

### LINKED ISSUE #507: Developer Tools Collection
Add a collection with helpful dev tools we've found over the internet. It'd be great to have a write up on it in a markdown file instead of plugging someone into app right away.

Perhaps having support for inline links in the description of the cards would be great. 

here is an example amazing tool https://sizzy.co/?url=https%3A%2F%2Fdeveloper.gov.bc.ca

patricksimonian:
## Summary
Added a card that speaks on the Sizzy web tools

Fixes #507

### DIFF
--- a/app-web/registry/developer-tools.json
+++ b/app-web/registry/developer-tools.json
@@ -12,7 +12,7 @@
       {
         "sourceType": "web",
         "sourceProperties": {
-          "resourceType": "Self-Service Tool",
+          "resourceType": "Self-Service Tools",
           "title": "OpenShift Developer Console",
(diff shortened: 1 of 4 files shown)


app-web/registry/developer-tools.json
```diff
@@ -29,6 +29,17 @@
           ]
         }
       },
+      {
+        "sourceType": "github",
+        "sourceProperties": {
+          "url": "https://github.com/bcgov/devhub-resources",
```
ShellyXueHan: just curious, why do you need the full url and the owner+repo ?

patricksimonian: i think this was an artifact of the very first version of devhub. I don't think it actually has any relevance anymore...

PR #612, opened 2019-05-08
Changed Collections to Topics and updated Unit Tests

lukegonis:
## Summary
Changed (UI part of this task) from Collections to Topics and updated Unit Tests, still need to change the variable/data structure for Collections. 

## Notable Changes 
-
-
-

### DIFF
--- a/app-web/src/constants/routes.js
+++ b/app-web/src/constants/routes.js
@@ -13,7 +13,7 @@ export const MAIN_NAV_ROUTES = {
   },
   COLLECTIONS: {
     to: '/collections',
-    text: 'Collections',
+    text: 'Topics',
   },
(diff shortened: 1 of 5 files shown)


app-web/src/constants/routes.js
```diff
@@ -13,7 +13,7 @@ export const MAIN_NAV_ROUTES = {
   },
   COLLECTIONS: {
     to: '/collections',
```
patricksimonian: does changing this to /topics break anything?

lukegonis: Yes, I believe I got a 404 when I tried that.

patricksimonian:
can you confirm this has been rebased on to master?

Issue #617, opened 2019-05-09
Update Topics view based on designs/mockups


patricksimonian:
@orinevares and @dlevineBC have some great designs that we have gone over for the next iteration of the Topics page.

Do either of you mind posting those images here as comments?

orinevares:
Sure, do you want the original sketch file as well? Whatever is easiest for you.

orinevares:
[Topics Mockuppdf.pdf](https://github.com/bcgov/devhub-app-web/files/3163295/Topics.Mockuppdf.pdf)

orinevares:
![Topics Mockup1pdf](https://user-images.githubusercontent.com/43620986/57473905-c4b4bd80-7245-11e9-8e49-c075c6cfc232.jpg)
![Topics Mockuppdf](https://user-images.githubusercontent.com/43620986/57473911-c67e8100-7245-11e9-9adf-58185fa0f1d9.jpg)

patricksimonian:
awesome. @DerekSgov let's extreme program this bad boy.

DerekSgov:
Lets do it @patricksimonian and thanks @orinevares for the mockups, they look great!

patricksimonian:
@orinevares does the sketch file show measurements and other css specs? If so, it may help. Also we do not have sketch so are their other ways to consume the file?

orinevares:
You could upload the sketch file into UX Pin. Otherwise I'd have to copy and paste out the individual layer CSS attributes from the sketch file on my computer.

orinevares:
Emailing you the sketch file just in case.

patricksimonian:
oh okay perfect, i'll take a look at that. I think i can upload it into Zeplin as well

patricksimonian:
@orinevares did you get those sketch files to me?

orinevares:
Yup should be in your inbox, but I'll send it over again now just in case.

PR #642, opened 2019-05-21
Begin integration of design tokens for better UI layout control

patricksimonian:
## Summary

Design Tokens are a way to maintain consistency among styles between components. Common styles are abstracted away into a constants file and used as template strings within the css-in-js components.
In addition I wrote up a quick doc explaining Design Tokens and how to use em'

This is in prep for #617

### DIFF
--- a/app-web/src/constants/designTokens.js
+++ b/app-web/src/constants/designTokens.js
@@ -4,6 +4,14 @@ export const STYLE_BASE = {
   borderWidth: 1,
 };
 
+// custom font size typing
+// CUSTOM_TYPE.sm => 16px;
+export const CUSTOM_TYPE = {
+  sm: `${STYLE_BASE}px`,
+  md: `${STYLE_BASE * 1.5}px`,
+  lg: `${STYLE_BASE * 2}px`,
+};
+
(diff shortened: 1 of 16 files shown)


app-web/src/constants/designTokens.js
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * returns an object of spacing units for things like padding/margin
+ * @param {Number} maxSpacingMultiple the maximum multiple of the spacing factor
+ * @param {Number} spacingFactor
+ * @returns {Object} => {'1x': '8px', '2x': '16px'};
+ */
+const getSpacingUnits = (maxSpacingMultiple, spacingFactor) => {
+  const spacingUnits = {};
+  for (let i = 1; i <= maxSpacingMultiple; i++) {
+    spacingUnits[`${i}x`] = `${i * spacingFactor}px`;
```
ShellyXueHan: so does this object get generate everytime you call SPACING? would it be easier if it's a function that returns like `20px` instead?

patricksimonian: I think `SPACING` is evaluated only once

PR #644, opened 2019-05-22
CSS fix for IE 11 text overflow on cards

DerekSgov:
## Summary
Just a one line CSS addition in the index.js file for the Card component


app-web/src/components/Cards/Card/index.js
```diff
@@ -50,6 +50,7 @@ export const CardWrapper = styled.article`
   box-shadow: 0 1px 2px 1px #00000026;
   display: flex;
   flex-flow: column nowrap;
+  overflow: hidden;
```
patricksimonian: for cards with extra long text, does this just cut the text off or can you scroll?

PR #660, opened 2019-05-23
Clearing up front page cards, without disturbing search functionality

### LINKED ISSUE #649: Clearing up card clutter on home page

DerekSgov:
## Summary
Changed the logic in the Index.js file (within pages folder) so that the cards are not shown by default on home page, while still maintaining the search functionality. Then updated the test files to match the new behaviour, and updated snapshots according.

close #649

### DIFF
--- a/app-web/__tests__/pages/Index.test.js
+++ b/app-web/__tests__/pages/Index.test.js
@@ -89,7 +89,7 @@ describe('Home Page', () => {
     expect(Alert).toBeInTheDocument();
   });
 
-  test('when there is an empty search the alert box does not show and all results show instead', () => {
+  test('when there is an empty search the alert box does not show and neither do cards', () => {
     queryString.parse.mockReturnValue({});
     const { container, rerender, queryByTestId, queryAllByTestId } = render(
       <ThemeProvider theme={theme}>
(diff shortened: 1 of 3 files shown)


patricksimonian:
@DerekSgov it looks like your commits are not authored by you. https://help.github.com/en/articles/setting-your-commit-email-address-in-git

DerekSgov:
oh weird, fixing that now - Thanks for the heads up

patricksimonian:
awesome!! Digging this, I'll review shortly

DerekSgov:
:) no rush, gotta rebase after the other pr is dealt with anyways

app-web/__tests__/pages/Index.test.js
```diff
@@ -89,7 +89,7 @@ describe('Home Page', () => {
     expect(Alert).toBeInTheDocument();
   });
 
-  test('when there is an empty search the alert box does not show and all results show instead', () => {
+  test('when there is an empty search the alert box does not show and neither do cards', () => {
```
patricksimonian: so by empty search do we mean that the path is `developer.gov.bc.ca/?q=` where a user search with an empty string?

DerekSgov: Yeah exactly

DerekSgov:
Thanks!!

app-web/__tests__/pages/Index.test.js
patricksimonian: I think we should reword this,  (this is initially my bad).
Something Like... when a blank search is entered, only the topics are visible

patricksimonian: something like that? 
We can then assert that card don't show
alerts don't show
and topics/collections do show.

DerekSgov: Good call, how about "When a blank search is entered, we can assert that cards and alerts don't show but topics/collections do"?

patricksimonian: sure but ditch the 'we can assert' part. the 'expect' function is the assertion

DerekSgov: gotcha, consider it done

patricksimonian: i should have bulletted my example from above:P

> something like that?
> We can then assert that card don't show
> alerts don't show
> and topics/collections do show.
> 
> what do you think?

DerekSgov: haha no worries, all good

PR #668, opened 2019-05-28
Removing duplicate cards and fixing duplication bug 

### LINKED ISSUE #665: Remove duplicate cards from Tools, Components and Documentation pages
and fix the duplication search bug where when "mobile" is searched, the card duplicates its self everytime it is searched

DerekSgov:
## Summary
-Removed duplicate cards
-Fixed card duplication bug when searching
-Created a test to check for this behaviour
-Update snapshot

## Notable Changes 
- For future reference: Resources in many collections will only be shown in one of their many contexts on the 'Tools', 'Components' and 'Documentation' pages

close #665

### DIFF
--- a/app-web/__tests__/pages/resourceType.test.js
+++ b/app-web/__tests__/pages/resourceType.test.js
@@ -14,6 +14,7 @@ import { TEST_IDS as NO_RESOURCE_TEST_IDS } from '../../src/components/UI/NoReso
 import { SEARCH as SEARCH_MESSAGES } from '../../src/messages';
 import DEFAULT_FILTERS, { FILTER_QUERY_PARAM } from '../../src/constants/filterGroups';
 import { TEST_IDS as FILTER_TEST_IDS } from '../../src/components/Filters/FilterGroup/FilterGroup';
+import { RESOURCE_TYPES } from '../../src/constants/ui';
 
 jest.mock('react-spinners', () => () => <div>loading</div>);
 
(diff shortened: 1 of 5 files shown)


app-web/__tests__/pages/resourceType.test.js
```diff
@@ -131,11 +132,36 @@ describe('Resource Type Template Page', () => {
...
+  test('Cards are not duplicating after multiple searches', () => {
+    queryString.parse.mockReturnValue({ q: 'foo' });
+    useSearch.mockReturnValue([{ id: SIPHON_NODES[0].id }]);
+    const { rerender, queryAllByText } = render(
+      <ThemeProvider theme={theme}>
+        <ResourceType {...props} />
+      </ThemeProvider>,
+    );
+
+    const startingNumCards = queryAllByText('Documentation').length;
```
patricksimonian: replace 'Documentation' with the `RESOURCE_TYPES.DOCUMENTATION` constant

app-web/__tests__/pages/resourceType.test.js
```diff
@@ -131,11 +132,36 @@ describe('Resource Type Template Page', () => {
     expect(Alert).toBeInTheDocument();
   });
 
+  test('Cards are not duplicating after multiple searches', () => {
```
patricksimonian: I'm liking this test however I feel like we could still run into this issue with the useSearch function misbehaving. This is because we are mocking out useSearch of course. Have you taken a look through the usesearch hook to ensure it will not create duplications?

app-web/registry/featured-cards.json
```diff
@@ -7,12 +7,13 @@
...
             "resourceType": "Self-Service Tools",
             "sourceProperties": {
+              "title": "Mobile Signing",
               "url": "https://signing-web-devhub-prod.pathfinder.gov.bc.ca"
             }
           },
           {
             "sourceType": "github",
-            "resourceType": "Documentation",
+            "resourceType": "Self-Service Tools",
```
patricksimonian: are you sure these are tools??

app-web/__tests__/pages/resourceType.test.js
DerekSgov: Done!

patricksimonian:
how does this look in dev?

DerekSgov:
its looking good to me! (and behaving as it should). what do you think?

patricksimonian:
hmm I'm thinking we should remove the building modern web apps card since that is going to be split up into two seperate cards ... what do you think?

DerekSgov:
Oh yeah thats a good point... Just from the featured cards topic?

PR #675, opened 2019-05-31
Adding Event resource type and eventbrite events to Topic sidebar

DerekSgov:
## Summary
-Adding Event resource type
-Adding Eventbrite events to the sidebar of the Community Enablers and Events topic

### DIFF
--- a/app-web/src/components/GithubTemplate/Navigation/Navigation.js
+++ b/app-web/src/components/GithubTemplate/Navigation/Navigation.js
@@ -73,6 +73,16 @@ const Navigation = ({ items }) => {
     );
   }
 
+  if (groupedResources[RESOURCE_TYPES.EVENTS]) {
+    groups.push(
+      <NavGroup
+        type={RESOURCE_TYPES.EVENTS}
+        key={RESOURCE_TYPES.EVENTS}
+        items={groupedResources[RESOURCE_TYPES.EVENTS]}
+      />,
+    );
+  }
+
   return <div>{groups}</div>;
 };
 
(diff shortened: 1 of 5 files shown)


app-web/src/components/GithubTemplate/Navigation/Navigation.js
```diff
@@ -73,6 +73,16 @@ const Navigation = ({ items }) => {
     );
   }
 
+  if (groupedResources[RESOURCE_TYPES.EVENTS]) {
```
ShellyXueHan: instead of pushing to the groups array in the if statement separately, why not just map it?

app-web/src/templates/SourceGithub_default.js
```diff
@@ -51,8 +53,25 @@ class SourceGithubMarkdownDefault extends React.Component {
       createElement: React.createElement,
       components: { 'component-preview': previewWithNode },
     }).Compiler;
+    let navigationItems = nav.items;
 
-    const navigation = <Navigation items={nav.items} />;
+    if (collection.name === 'Community Enablers and Events') {
```
ShellyXueHan: better to use TOPICS.COMMUNITY_ENABLERS_AND_EVENTS from constants

app-web/src/templates/SourceGithub_default.js
```diff
@@ -51,8 +53,25 @@ class SourceGithubMarkdownDefault extends React.Component {
...
+      const eventbriteNavItems = flattenGatsbyGraphQL(communityEvents.edges);
+      const currentEvents = eventbriteNavItems
+        .filter(e => e.start.daysFromNow <= 0)
+        .map(event => ({
```
ShellyXueHan: do you also need to map the organization, logo, venue to the Event props?

PR #705, opened 2019-06-10
Redesigning event page and cards

DerekSgov:
## Summary
- Redesign event page
- Redesign event card when its an eventbrite event
- updating tests and snapshots
- bit of refactoring on the eventbrite siphon

### DIFF
--- a/app-web/src/components/Cards/Card/Card.js
+++ b/app-web/src/components/Cards/Card/Card.js
@@ -30,10 +30,15 @@ import {
   CardTitle,
   CardWrapper,
   DecorativeBar,
+  EventInfoDiv,
+  EventDate,
+  EventContainer,
+  EventImageWrapper,
 } from './index';
 import Aux from '../../../hoc/auxillary';
 
 import { RESOURCE_TYPES_LIST } from '../../../constants/ui';
+import EventLogo from '../../Event/Logo';
 
 const Card = ({ type, title, description, image, link, ...rest }) => {
   let isExternal = !!validUrl.isWebUri(link);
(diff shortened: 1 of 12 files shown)


app-web/src/components/Cards/Card/Card.js
```diff
@@ -45,7 +50,31 @@ const Card = ({ type, title, description, image, link, ...rest }) => {
     </CardDescription>
   );
 
-  if (image && description) {
+  //first check if its an eventbrite event, as in our resolver in gatsby-node.js -> image is set to "eventbrite"
+  if (image === 'eventbrite' && description) {
+    cardBody = (
+      <Aux>
+        <CardDescription title={description} clamp={3} tagName="p">
```
patricksimonian: Instead of making the card component smarter, (it has to keep track of event brite events...and than meetups, and then yammer) let's make it dumber. There is the `renderProps` feature in react which may work well for this. If there are render props, don't render a card body, just render the `renderProps`.

app-web/src/templates/events.js
```diff
@@ -83,53 +87,35 @@ export const EventsPage = ({ data: { allEventbriteEvents, allDevhubCollection }
...
-            community.
-          </CollectionDescription>
-          <ContainerBoarder>
-            <CardCarousel resources={communityCards} />
-            <LinkContainer>
-              <ChevronLink to={getFirstNonExternalResource(communityCards)}>View</ChevronLink>
-            </LinkContainer>
-          </ContainerBoarder>
-        </Aux>
+        <span>&nbsp;&nbsp;</span>
```
patricksimonian: what's this for?

Issue #518, opened 2019-03-28
As a developer I would like to be able to use ANY gatsby source/transformer plugin as data source for the devhub


patricksimonian:
The idea is that we want to be able to utilize any gatsby source plugin and via an abstract class factory implement a standard interface that all these plugins would map too. This would be via a transformer plugin that knows how to map a source plugin/ other transformers

patricksimonian:
After a lot of discovery, I've found that leveraging the features of gatsby itself are more than enough to accomplish 90% of our functionality (if not more). The open PR #719, is left open as leveraging 3rd party plugins and making the connections between them has introduced some dependancies that need to be worked on.

These dependancies are:

- `gatsby-source-github-raw` anchor tag transformation is required so that any links inside of markdown files are appropriately pointing to devhub pages when they exist
- `gatsby-plugin-elastic-lunr` search engine needs to be reconfigured to work with pure markdown files

## So what has been done?

Extensive experimentation and proof of concepts (all documented, and even illustrated!) on how to connect 3rd party nodes together.

PR #714, opened 2019-06-13
add validators functions to be used on node field creations

patricksimonian:
## Summary
This pr is an effort for the normalization of many source plugins. I'm taking a stab at doing preliminary normalization through the `createNodefields` function. Before that can be done we need a set of validators to run functions against when nodes are created. Hence this pr.


patricksimonian:
#518

PR #736, opened 2019-06-24
Adding more introductory OpenShift materials for those waiting to take the OpenShift 101

DerekSgov:
## Summary
-Adding resources to registry
-Adding topic to home page so it can easily be found

### DIFF
--- a/app-web/src/utils/helpers.js
+++ b/app-web/src/utils/helpers.js
@@ -165,21 +165,26 @@ export const isFilterLonely = filters => {
  *
  * @param {Array} topics the raw topic data (aka collections) as taken from the graphql query
  * @param {Object} topics[ind] {node: {..., name: {string }}}
- * @returns {Array} the sorted topics with Featured Cards and Design System first!
+ * @returns {Array} the sorted topics with Featured Cards, Design System and Getting Started on the DevOps Platform first!
  */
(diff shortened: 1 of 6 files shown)


app-web/src/utils/helpers.js
```diff
@@ -165,21 +165,26 @@ export const isFilterLonely = filters => {
...
+    topic => topic.node.name === TOPICS.GETTING_STARTED_ON_THE_DEVOPS_PLATFORM,
   );
 
-  const sortedTopics = topicsWithoutDSAndFC.sort((topicA, topicB) => {
+  const designSystem = { ...topics[designIndex] };
+  const featuredResources = { ...topics[featuredIndex] };
+  const devOpsPlatform = { ...topics[devOpsPlatformIndex] };
+  // remove selected topics
+  const topicsWithoutSelectedTopics = topics
+    .filter(topic => topic.node.name !== TOPICS.FEATURED_RESOURCES)
```
patricksimonian: this _could_ be combined into one filter, prevents the loop from running 3x. But I think if we were talking Big (O) this would all still be O(n) right?

DerekSgov: yeah still O(n), but it will look cleaner that way

DerekSgov:
Probably a good thing to do, will do now!

PR #765, opened 2019-07-03
Imp/search index

### LINKED ISSUE #742: Devhub V3: update gatsby lunr plugin search index to reference markdown nodes/events
The search index should have a single source of truth for documentation resources which will be a MarkdownRemark gatsby node. 

The great thing with this is that all MarkdownRemark nodes have a parent that they will reference. 

In addition we should index event nodes and have a way of surfacing them in parallel with markdown nodes during searches (these are the trials and tribulations of working with static content :) ) 

As a node any work should be branched from `imp/gatsby-source-github-raw-integration` and merged back into that branch! sorry for the ugly name. 

> why using an integration branch? Technically imp/gatsby-source-github-raw-integration can be merged into master as its features are not being used within the page components and it supports backwards compatibility with siphon, however because this is such a __large__ change, i felt throwing in code to production/master that may be stripped out later was not such a good deal

DerekSgov:
## Summary
-Adding MarkdownRemark, Eventbrite and Meetup nodes to the search index
-Linking them up in search
-Fixing css formatting issue
-Updating tests and snapshots

quick note - as of now, there are many results not being shown to prevent card barf of not very relevant search results. The next step will be to find a way to pick relevant search results from the index

close #742

### DIFF
--- a/app-web/src/pages/index.js
+++ b/app-web/src/pages/index.js
@@ -23,6 +23,8 @@ import {
 import { isQueryEmpty } from '../utils/search';
 import { SEARCH_QUERY_PARAM } from '../constants/search';
 import { SPACING } from '../constants/designTokens';
+import uniqBy from 'lodash/uniqBy';
+import { formatEvents } from '../templates/events';
 
 const Main = styled.main`
   margin-bottom: ${SPACING['1x']};
@@ -87,6 +89,7 @@ export const Index = ({
   data: {
     allDevhubCollection,
     allDevhubSiphon,
+    allEventbriteEvents,
     siteSearchIndex: { index },
   },
   location,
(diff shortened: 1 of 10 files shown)


app-web/src/pages/index.js
```diff
@@ -103,14 +107,31 @@ export const Index = ({
...
+  const currentEvents = formatEvents(allEvents.filter(e => e.start.daysFromNow <= 0));
+  const allMeetups = formatMeetUps(
+    flattenGatsbyGraphQL(allMeetupGroup.edges).flatMap(meetups => {
+      return meetups.childrenMeetupEvent;
+    }),
+  );
+  const currentMeetups = allMeetups.filter(e => e.start.daysFromNow <= 0);
+  const eventsAndMeetups = currentEvents.concat(currentMeetups);
+
+  //const currentEvents = formatEvents(events.filter(e => e.start.daysFromNow <= 0));
```
patricksimonian: commented out code

app-web/src/utils/hooks.js
```diff
@@ -26,13 +26,22 @@ import isEqual from 'lodash/isEqual';
...
       if (Index === null) {
         setIndex(ElasticLunr.load(staticIndex));
       } else {
-        const searchResults = Index.search(query, { expand: true }).map(({ ref }) =>
+        const searchResults = Index.search(query, { fields: {
+          title: { boost: 4 },
+          content: { boost: 1 },
+          description: { boost: 1 },
+          collectionName: { boost: 2 },
+          labels: { boost: 2 },
```
patricksimonian: lets give labels a better boost since it is a field that authors dictate themselves

PR #779, opened 2019-07-08
Only build component if there are changes associated with current Jenkinsfile

### LINKED ISSUE #768: Build should only trigger when changes related to their "component" are detected.
Currently, we are triggering builds for all components in a repo, regardless of whether there are changes to a particular component.  We need to tweak the code that checks for changes in each Jenkinsfile, or find another mechanism to do the same.  This will reduce the number of builds we have running and resource utilization.

sheaphillips:
Closes #768.

We should not trigger a build for a component in a repo when changes pushed are not associated with it.

This PR contains modifications to the Jenkinsfiles for both devhub-app and Matomo to stop them from building unless changes to "their" files are contained in a triggering commit.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,9 @@
 pipeline {
     agent none
+    environment {
+        COMPONENT_HOME = '/'
+        BUILD_TRIGGER_EXCLUDES = '^.jenkins/||^matomo/'
+    }
     options {
         disableResume()
     }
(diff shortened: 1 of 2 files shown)


sheaphillips:
This is a couple of pipeline only changes that will reduce our resource utilization and reduce/eliminate unnecessary builds.  Tested and confirmed working.

sheaphillips:
Note: Once reviewed+approved, this can be merged and closed - we don't want/need to promote associated builds to Prod.

matomo/Jenkinsfile
```diff
@@ -1,5 +1,10 @@
 pipeline {
     agent none
+    environment {
+            COMPONENT_NAME = 'Matomo Analytics'
+            COMPONENT_HOME = 'matomo'
+            BUILD_TRIGGER_INCLUDES = '^matomo/'
```
patricksimonian: @cvarjao I think this would be a great feature for the bcdk to have this included as apart of the pipeline.

Jenkinsfile
```diff
@@ -1,5 +1,10 @@
 pipeline {
     agent none
+    environment {
+        COMPONENT_NAME = 'DevHub web app'
+        COMPONENT_HOME = '/'
+        BUILD_TRIGGER_EXCLUDES = "^.jenkins/\\|^matomo/"
```
patricksimonian: can this use an INCLUDES regex over the excludes? is it because it is a root project jenkinsfile?

sheaphillips: Yes - in this case, it requires an excludes because there is/was code in the root that related to DevHub.  Once the folder restructure is in place it can just be an 'include'.  

We can definitely come up with something better for this, but IMO it ultimately belongs outside the Jenkinsfile and integrated/generated by bcdk.

PR #782, opened 2019-07-08
Work towards speeding up builds via use of incremental build.

sheaphillips:
## Summary
Builds are too slow.  We want to leverage OpenShift incremental builds if we can to speed things up.

## Notable Changes 

- removed imagesream/bc and associated files for project  "base" image
- enabled incremental builds in node-based bc.


sheaphillips:
This is just the refactoring of the build artifacts to no longer use the "base" image, plus adding the "incremental" setting.  Inconclusive ATM if this is improving things, but it's the right config to have.   We can do some follow-on work to determine if it's optimal.  

Note this doesn't need to be promoted, so proposing just merging/closing once reviewed + approved.  

Note also - snyk is confused due to files having been removed and has a "false negative" check..

PR #816, opened 2019-07-12
Integrate resource type via fields

patricksimonian:


app-web/src/pages/index.js
```diff
@@ -72,6 +72,8 @@ const getResourcePreviews = (resources, results = []) => {
   }
 
   resourcesToGroup = getUniqueResources(resourcesToGroup);
+
+
```
ShellyXueHan: miss anything here?

Issue #781, opened 2019-07-08
Evaluate/Implement incremental build for devhub web app.


sheaphillips:
In theory, incremental build should speed up build times for the devhub web app.  We should dig in and experiment to confirm. If it does, then we should definitely start using this approach.

PR #829, opened 2019-07-17
Adding s2i scripts to enable incremental builds

sheaphillips:
The builder image that we are currently using in the cluster does not support incremental builds as-is.  This PR adds s2i scripts that will cause it to support incremental builds, with the goal of speeding up build performance in the cluster.

Also, this PR is adding a utility script to perform s2i build locally.


sheaphillips:
Closes #781

sheaphillips:
*Note*: there's something weird going on with the checks/actions bit, but I confirmed that the workflow steps worked.   going ahead and merging.

PR #834, opened 2019-07-19
Search Result Total PillBox

DerekSgov:
## Summary
- when searching on the home page, there is now a pill box to show the total number of results


app-web/src/components/Search/SearchPills.js
```diff
@@ -25,19 +25,21 @@ const SearchPill = styled(Pill)`
 const PillContainer = styled.div`
   display: flex;
   flex-flow: row wrap;
+  float: left;
```
patricksimonian: why are ya floating left?

PR #843, opened 2019-07-23
Bug/search results fix

DerekSgov:
## Summary
- The issue was that the 'No resources found :(" banner was often not being shown when it should be, this was because things like past events (on the home page search) or results of the wrong resource type (resourceType pages) where coming back in the results
- On each the home and resourceType pages, I created a method to filter out unwanted results so that the logic and later code now returns proper behaviour

### DIFF
--- a/app-web/src/utils/helpers.js
+++ b/app-web/src/utils/helpers.js
@@ -63,6 +63,37 @@ export const getTextAndLink = (resourceType, resourcesByType) => {
 };
 
 /**
+ * Removes results of the wrong resource type from the search results
+ * @param {array} results the search results returned from our index
+ * @param {string} resourceTypeConst the given resource type for this page
+ * @param {array} resources all the resources on the site
+ */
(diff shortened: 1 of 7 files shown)


app-web/src/utils/helpers.js
```diff
@@ -62,6 +62,37 @@ export const getTextAndLink = (resourceType, resourcesByType) => {
...
+/**
+ * Unwanted results are things like past events which are returned in our index but we do not want to show
+ * having these events in our results can mess up some of our logic later down the line (i.e for no results found)
+ * returns results, but without any past events etc
+ * @param {array} results the search results returned from our index
+ * @param {array} allEventsAndMeetups all the events and meetups on the site
+ * @param {array} currentEventsAndMeetups all the current events and meetups on the site
+ */
+export const removeUnwantedResults = (results, allEventsAndMeetups, currentEventsAndMeetups) => {
+  let allIDs = allEventsAndMeetups.map(event => event.siphon.id);
```
patricksimonian: we should drop the siphon reference:)

PR #889, opened 2019-08-02
Imp/apollo client component

DerekSgov:


app-web/src/components/Search/Search.js
```diff
@@ -0,0 +1,50 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+
+Created by Derek Siemens
+*/
+
+import React from 'react';
+import { Query } from 'react-apollo';
+//run time query
+import gql from 'graphql-tag';
```
patricksimonian: can we use gatsby native `graphql`?

patricksimonian: instead of `gql`?

`import {graphql} from 'gatsby'`

app-web/wrapWithProvider.js
```diff
@@ -20,9 +20,21 @@ import React from 'react';
...
 import { AuthProvider } from './src/AuthContext';
 import theme from './theme';
+import { ApolloClient, InMemoryCache } from 'apollo-boost';
+import { ApolloProvider } from 'react-apollo';
+import { createHttpLink } from 'apollo-link-http';
+
+const cache = new InMemoryCache();
+//this needs to be an enviroment variable later on......
+const client = new ApolloClient({
+  link: createHttpLink({ uri: 'https://rocketgate-dev-devhub-dev.pathfinder.gov.bc.ca' }),
```
patricksimonian: this needs to come in as a config similar to how `auth.js` does it. It will either be an environment variable or a caddy template string. Checkout `config.js`

PR #912, opened 2019-08-16
Sort search results by relevance, and show result count by resource type with interaction

DerekSgov:
## Summary
- Search results are no longer grouped by resource type but now by relevance. The resource type metadata can be seen above the results with their respecting result counts
- When clicked, the resource type pills now filter the results being shown according to the given resource type
- click on = filter, click off = all results back
- RocketChat Icon can now be seen even if not logged in, but when the user isnt logged in - it is faded

### DIFF
--- a/app-web/src/components/Home/ResourcePreview.js
+++ b/app-web/src/components/Home/ResourcePreview.js
@@ -18,13 +18,14 @@ Created by Patrick Simonian
 import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
+import css from '@emotion/css';
 
 import { EMOTION_BOOTSTRAP_BREAKPOINTS } from '../../constants/designTokens';
 import { ChevronLink } from '../UI/Link';
 import { Container, LinkContainer } from './index';
 import Card from '../Cards/Card/Card';
 import Pill from '../UI/Pill';
-import css from '@emotion/css';
+import { RESOURCE_TYPES } from '../../constants/ui';
 
 export const CardWrapper = styled.div`
   margin: 6px 9px;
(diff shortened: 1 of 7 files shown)


app-web/src/components/Home/ResourcePreview.js
```diff
@@ -15,14 +15,16 @@ limitations under the License.
...
 import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
 
 import { EMOTION_BOOTSTRAP_BREAKPOINTS } from '../../constants/designTokens';
 import { ChevronLink } from '../UI/Link';
-import { Container, Title, StyledLink, LinkContainer } from './index';
+import { Container, LinkContainer } from './index';
 import Card from '../Cards/Card/Card';
+import Pill from '../UI/Pill';
+import css from '@emotion/css';
```
patricksimonian: for consistency, let's move this to line 21

app-web/src/components/Home/ResourcePreview.js
```diff
@@ -38,43 +40,179 @@ export const ResourceContainer = styled.div`
...
-        <CardWrapper key={r.id}>
-          <Card
-            type={r.fields.resourceType}
-            title={r.fields.title}
-            description={r.fields.description}
-            image={r.fields.image}
-            link={r.fields.standAlonePath}
-            event={r}
+export const ResourcePreview = ({ title, link, resources, filters, amountToShow, seeMore }) => {
+  let [showCount, updateCount] = useState(amountToShow);
```
patricksimonian: really good, the convention for naming your 'setter' function with `useState` is 'setCount', 'setSeeMore'

app-web/src/components/Home/ResourcePreview.js
```diff
@@ -38,43 +40,179 @@ export const ResourceContainer = styled.div`
...
+    updateSeeMore(seeMore);
+  };
+
+  let pills;
+  //Filters will be mapped into pills displaying result count for that particular resourcetype
+  //These pills are interactive and filter results when clicked
+  if (filters) {
+    //No people resource type Icon rn
+    pills = filters.map(filter => {
+      if (filter.name !== 'People') {
```
patricksimonian: is that a resource type?

app-web/src/components/Home/ResourcePreview.js
```diff
@@ -38,43 +40,179 @@ export const ResourceContainer = styled.div`
...
+        //formats the text correctly for different cases
+        let iconLabel =
+          filter.counter > 1 || filter.counter === 0
```
patricksimonian: could you just do if `filter.count === 1 ?`

app-web/src/components/Home/ResourcePreview.js
```diff
@@ -38,43 +40,179 @@ export const ResourceContainer = styled.div`
...
-      <ChevronLink to={link.to}>{link.text}</ChevronLink>
-    </LinkContainer>
-  </Container>
-);
+        );
+      }
+      return '';
+    });
+    //Add final pill to the start of the list
+    pills.unshift(
```
patricksimonian: hmm i'd avoid using unshift. I think it won't cause any issues here but it is an object mutating method. Since we are typically using concat/filter to add and remove elements. Let's stick to that? What do you think?

app-web/src/components/Home/ResourcePreview.js
```diff
@@ -38,43 +40,179 @@ export const ResourceContainer = styled.div`
...
+              description={r.fields.description}
+              image={r.fields.image}
+              link={r.fields.standAlonePath}
+              event={r}
+            />
+          </CardWrapper>
+        ))}
+      </ResourceContainer>
+      <LinkContainer>
+        {seeMoreResults && resourcesToShow.length > showCount && (
```
patricksimonian: nice!

PR #921, opened 2019-08-21
fix an issue where the rc logo would not anchor in safari

### LINKED ISSUE #899: bug for rocket chat icon on safari
when using Devhub on safari while logged in, after making a search, going to an external site and hitting back on your browser - the functionality of the rocketchat button no longer works

patricksimonian:
This does not fix issue #899. Unfortunately, I am unable to do discern what the root cause is at this stage. My guess is that if the URL already matches the contents of the anchor tag's href attribute, clicking on the anchor tag does not work as expected (it doesn't scroll to the anchored element)

### DIFF
--- a/app-web/src/components/Search/SearchSources.js
+++ b/app-web/src/components/Search/SearchSources.js
@@ -25,9 +25,9 @@ export const SearchSources = ({ rocketchat }) => {
   }
   return (
     <SearchSourcesContainer data-testid={TEST_IDS.container}>
-      <Link to={'#rocketChat'}>
+      <a href="#rocketChat">
         {authenticated && <RCButton {...rcProps} title="toggle rocket chat search results" />}
-      </Link>
+      </a>
     </SearchSourcesContainer>
   );
 };


patricksimonian:
Unfortunately, my localhost testing gave me a false positive that this issue had been resolved. There are javascript based ways we can solve this issue. @sheaphillips what do you think?

sheaphillips:
Does this affect regular anchors as well or just ones when they’re is a query string plus anchor?

patricksimonian:
> Does this affect regular anchors as well or just ones when they’re is a query string plus anchor?
> […](#)
> Sent from my iPhone
> On Aug 22, 2019, at 9:51 AM, Patrick Simonian ***@***.***> wrote: Unfortunately, my localhost testing gave me a false positive that this issue had been resolved. There are javascript based ways we can solve this issue. @sheaphillips what do you think? — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or mute the thread.

Unfortunately, it is any anchor with a `#hash` in it. From my debugging, I've narrowed it down to if a page was navigated to with a `#hash` subsequent clicks to the anchor that generated that link do not work.

patricksimonian:
closing this issue in favor of issue #941 which will address this ui bug

PR #926, opened 2019-08-22
Imp/909 dynamic featured topic

### LINKED ISSUE #909: Make the "Featured" topic dynamic
Make the "Featured" topic dynamic so that a file is not required in the registry for that topic, perhaps take inspiration from how the "Popular" topic is done

--This is also as preparation for the community contributed topic

patricksimonian:
## Summary
Convert featured topic to render dynamically as apart of a function (similar to how the popular topic works)

This allows us to not have to maintain a registry item (which would duplicate nodes) for a featured topic. 

Fixes #909


app-web/src/pages/index.js
```diff
@@ -47,6 +53,7 @@ const Main = styled.main`
  * @param {Boolean} searchResultsExist
  */
 const getTopicPreviews = (topics, searchResultsExist) => {
+  console.log(topics);
```
DerekSgov: Probably want to delete the console.log()

PR #947, opened 2019-09-04
Update redirect uri to include search query

### LINKED ISSUE #940: Enable search persistence after login
A user searches for something before logging in, then realize they need to login (to see RC results for example), they then log in, and are sent to the home page instead of to the search results page they were just on. They should be sent to the search results page they were just on.

patricksimonian:
## Summary
Fixes #940 where bookmarking a search query would not work because the implicit auth redirect uri would ignore the query params

### DIFF
--- a/app-web/src/auth.js
+++ b/app-web/src/auth.js
@@ -19,6 +19,8 @@ export const createIam = () => {
 
   if (iam === null) {
     iam = new ImplicitAuthManager(config);
+    window._iam = iam;
+
     Object.freeze(iam);
   }
   return iam;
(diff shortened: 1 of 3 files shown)


app-web/src/auth.js
```diff
@@ -8,14 +8,19 @@ let iam = null;
...
     baseURL: SSO_BASE_URL,
     clientId: SSO_CLIENT_ID,
     realmName: SSO_REALM_NAME,
-    redirectURI: window.location.origin + window.location.pathname,
+    redirectURI: origin + pathname + search,
   };
+
   if (iam === null) {
     iam = new ImplicitAuthManager(config);
+    window._iam = iam;
```
w8896699: Remove global.

PR #950, opened 2019-09-06
Replace rocket gate apollo client with search gate

### LINKED ISSUE #907: Adapt search client to new (Search.Gate) API and endpoint

patricksimonian:
## Summary
As described by issue #907, we are leveraging search gate to perform our runtime graphql searches. This returns payloads from different search sources in a standard schema. 

This pr integrates search gate and modifies any infra code that referenced Rocketgate with Search gate

Reference repositories for the two can be found here
https://github.com/bcgov/searchgate
https://github.com/bcgov/rocketgate

fixes #907


patricksimonian:
looks like there is a few ui bugs.

- when searching, static resources load before dynamic ones. You will see the search result count jump
- static search results that or not relevant to a new search still show up when searching rocket chat
   - navigate to dev instance of devhub
   - search react
   - search openshift

PR #951, opened 2019-09-09
gate search gate with auth

patricksimonian:
## Summary
search gate only shows results when authenticated


w8896699:
Merge it so I can rebase and sent my pr

PR #956, opened 2019-09-11
Cancel placeholder for search input and update snapshot

w8896699:
Instead of using placeholder, we are now using searching query from last time as value after click search(Intimate what other search engine is doing).

### DIFF
--- a/app-web/src/components/Search/index.js
+++ b/app-web/src/components/Search/index.js
@@ -49,7 +49,6 @@ export const Search = ({ onSearch, searchOnEnter, inputConfig, query, ...rest })
 
   const search = () => {
     onSearch(terms);
-    setTerms('');
   };
 
   return (
--- a/app-web/src/utils/helpers.js
+++ b/app-web/src/utils/helpers.js
@@ -61,11 +61,12 @@ export const getTextAndLink = (resourceType, resourcesByType) => {
   return textAndPath;
 };
 
-/** return a combine text result or results depens on how much search results we have
+/** pluralize return string (results) if numOfResults is > 1. And return not search result if
+ * numOfResults === 0 example output: 10 Results / 1 Result / No search Result
  * @param {num} numOfResults
  * @returns {string}
  */
-export const getSearchResultLable = numOfResults => {
+export const getSearchResultLabel = numOfResults => {
   let resultLabel;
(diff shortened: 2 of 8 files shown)


app-web/src/components/Search/index.js
```diff
@@ -59,7 +59,7 @@ export const Search = ({ onSearch, searchOnEnter, inputConfig, query, ...rest })
         type="text"
         query={query}
         aria-label={ARIA_LABEL_SEARCH_INPUT}
-        value={terms}
+        value={terms.useState}
```
patricksimonian: where `.useState` coming from? `terms` should be fine on its own, I think what you are looking for is to initially set terms with a query if its available. In that case, you could use a `useEffect(() => {...}, [])` to accomplish this.

app-web/src/components/Search/index.js
```diff
@@ -49,9 +49,10 @@ export const Search = ({ onSearch, searchOnEnter, inputConfig, query, ...rest })
 
   const search = () => {
     onSearch(terms);
-    setTerms('');
   };
 
+  const handleFocus = event => event.target.select();
```
patricksimonian: what does this do?

app-web/src/utils/helpers.js
```diff
@@ -61,11 +61,13 @@ export const getTextAndLink = (resourceType, resourcesByType) => {
   return textAndPath;
 };
 
-/** return a combine text result or results depens on how much search results we have
+/** return a combine text result or results depends on how much search results we have.
```
patricksimonian: i'd think this top line of documentation is redundant

app-web/src/components/Search/index.js
w8896699: This will allow single click and select all text in input box

patricksimonian: hmm I'm not sure if that will be an effect we will want 100% of the time when searching. Just my opinion. @sheaphillips ?

I feel like if you have 'searched' than a single click highlight would be an awesome UX improvement. If it happens always, it prevents users from 'fixing typos'

patricksimonian: Also as a tip, i'd add a tiny comment above that event handler that explains exactly what the function does  :)

PR #957, opened 2019-09-12
Adding README (trivial change to trigger pipeline).

sheaphillips:
## Summary
Quick Summary Here

## Notable Changes 
-
-
-


patricksimonian:
as far as the cypress e2e test failure. I think i narrowed down the issue to cypress using Electron as a default headless browser. There is an image of cypress that contains chrome but I'm not sure how to leverage it within our form of github actions. Github did notify me that they are going away from their Jenkins-like DSL to a YAML format, maybe at that point it will allow us to select an image to use as an agent...

Issue #939, opened 2019-09-04
Provide 'info' box to let users know they can get more search results when logged in


patricksimonian:
Currently, when searching on the Devhub it is not immediately obvious that you need to be logged in to leverage rocket chat results.

Let's implement an alert style box (similar to the info alert box from bootstrap) that will provide unauthenticated users with a 'tip' about logging in to get more search results. 

Something like: "You can view search results from applications like Rocket.Chat or Github when logged in"

PR #958, opened 2019-09-12
Add a alert for user to let then login to view more info

w8896699:
> Added an info box that provides the user with a tip that they can view more search results from github/rocketchat when logged in. Once acknowledged the message does not show up for the remainder of the application lifecycle. 

The alert will not show up again once it is close, and will only show if user is not logged in and search something
#939 
-
-
-

### DIFF
--- a/app-web/__tests__/components/Masthead.test.js
+++ b/app-web/__tests__/components/Masthead.test.js
@@ -1,0 +1,62 @@
+import React from 'react';
+import { render, cleanup, fireEvent } from 'react-testing-library';
+import Masthead, { TEST_IDS } from '../../src/components/Home/Masthead';
+import { ThemeProvider } from 'emotion-theming';
+import theme from '../../theme';
+import { useAuthenticated } from '../../src/utils/hooks';
+import 'jest-dom/extend-expect';
+
+jest.mock('../../src/utils/hooks');
+describe('Masthead Component', () => {
+  afterEach(cleanup);
+
+  it('It shows an alert window when a search has happened', () => {
+    useAuthenticated.mockReturnValue({ authenticated: false });
+
+    const { queryByTestId, rerender } = render(
+      <ThemeProvider theme={theme}>
+        <Masthead query={'openshift'} searchSourcesLoading={true} resultCount={12} />
+      </ThemeProvider>,
+    );
+    const alert = queryByTestId(TEST_IDS.alertBox);
+    expect(alert).toBeInTheDocument();
+
+    rerender(
+      <ThemeProvider theme={theme}>
+        <Masthead query={'devop'} searchSourcesLoading={true} resultCount={0} />
+      </ThemeProvider>,
+    );
+    expect(alert).toBeInTheDocument();
+
+    const closebtn = alert.querySelector('button');
+    fireEvent.click(closebtn);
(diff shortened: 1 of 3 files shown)


app-web/src/components/Search/index.js
```diff
@@ -38,18 +41,29 @@ const Container = styled.div`
...
+  margin: 10px auto;
+`;
 export const TEST_IDS = {
   input: 'searchbar-input',
   button: 'searchbar-button',
 };
 
 export const Search = ({ onSearch, searchOnEnter, inputConfig, query, ...rest }) => {
   const [terms, setTerms] = useState('');
+  const { authenticated } = useAuthenticated();
```
patricksimonian: The alert box should live outside of the search component. This will have it appear on the resource type pages /components, /documentation etc.

w8896699: Thanks, I have rebased and move this to masthead.js

app-web/src/components/Home/Masthead.js
```diff
@@ -28,6 +28,8 @@ import Search from '../Search';
 // localizations
```
patricksimonian: remove this line, it not longer makes sense

app-web/src/components/Home/Masthead.js
```diff
@@ -101,6 +116,11 @@ export const Masthead = ({ query, resultCount, searchSourcesLoading }) => {
         />
         <IconDiv>{!searchSourcesLoading && query && resultCount > 0 && <SearchSources />}</IconDiv>
       </SearchContainer>
+      {!authenticated && alertHasBeenAcknowledged && (
```
patricksimonian: let's write a unit test in the masthead component for this.

app-web/src/components/Home/Masthead.js
```diff
@@ -60,6 +66,15 @@ const IconDiv = styled.div`
 `;
 
 export const Masthead = ({ query, resultCount, searchSourcesLoading }) => {
+  const { authenticated } = useAuthenticated();
+  let visible = !(query === '');
```
patricksimonian: I feel this could be simplified to something that is less verbose, in addition, this is a stateful variable so let's use `useState`

app-web/__tests__/components/Masthead.test.js
```diff
@@ -0,0 +1,71 @@
...
+    // if i perform another search (re render) after cancelling, the alert window does not show
+    rerender(
+      <ThemeProvider theme={theme}>
+        <Masthead query={'happy'} searchSourcesLoading={true} resultCount={0} />
+      </ThemeProvider>,
+    );
+
+    expect(alert).not.toBeInTheDocument();
+
+    // I will see the alert window
```
patricksimonian: Awesome!! let's delete these to comments since they are not needed

app-web/__tests__/components/Masthead.test.js
```diff
@@ -0,0 +1,71 @@
...
+    // if i perform another search the alert still exists
+  });
+
+  it('The alert box does not show when authenticated', () => {
+    // If the user is not authenticated, i do not see the alert window
```
patricksimonian: if the user is authenticated, they will not see the alert window

PR #960, opened 2019-09-13
Refactor Card Component to allow for dynamic headers/body and add storybook

patricksimonian:
## Summary
As described above, the card component has been greatly decomplicated*. It is now a set of composable sub-components. 

I also added storybook js to demonstrate/document the usage of these sub-components

### DIFF
--- a/app-web/src/components/Card/Card.js
+++ b/app-web/src/components/Card/Card.js
@@ -123,19 +123,15 @@ export const Card = ({
 
   return (
     <BaseCard resourceType={resourceType} link={link} {...rest}>
-      {renderBody ? (
-        renderBody()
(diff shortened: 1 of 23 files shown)


app-web/src/components/Card/Card.js
```diff
@@ -0,0 +1,219 @@
...
+  description,
+  image,
+  link,
+  renderBody,
+  renderHeader,
+  ...rest
+}) => {
+  let isExternal = !!validUrl.isWebUri(link);
+  let cardBody = null;
+  let inferredVariant = 'basic';
```
ShellyXueHan: use variants.basic instead?

PR #994, opened 2019-09-23
fix bug for search input bar

w8896699:
## Summary
add useeffect to change state if query is empty
fixed#992

### DIFF
--- a/app-web/src/components/Search/index.js
+++ b/app-web/src/components/Search/index.js
@@ -15,7 +15,7 @@ limitations under the License.
 
 Created by Patrick Simonian
 */
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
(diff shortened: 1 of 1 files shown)


app-web/src/components/Search/index.js
```diff
@@ -45,12 +45,15 @@ export const TEST_IDS = {
 };
 
 export const Search = ({ onSearch, searchOnEnter, inputConfig, query, ...rest }) => {
-  const [terms, setTerms] = useState('');
-
+  const [terms, setTerms] = useState(query);
```
patricksimonian: im not sure if we need to preset this state with query

w8896699: I put query here because I want query still stay in the input box after refresh current page

PR #1042, opened 2019-10-15
Journey Integration

patricksimonian:
## Summary
This will serve as an integration branch. There is still some work need to be done before this feature is released

- [x] render journeys on homepage
- [x] render journey 'detail' page
- [x] render journey list page


w8896699:
This pr seems has a test failing > index.test.js

PR #1078, opened 2019-11-04
Update nr repo name in registry

patricksimonian:
## Summary
The repo `nr-document-generation-showcase` was changed to `document-generation-showcase` which was causing failures to build. 

@sheaphillips should we make the devhub fault tolerant to this type of issue?


sheaphillips:
I think we definitely should be resilient to this...ideally we would
trigger warnings (vs failures) and get notifications

PR #1083, opened 2019-11-05
fix curl request

patricksimonian:
## Summary
update invalid curl request


patricksimonian:
right I can address that uhoh. It is apart of the git history now though 👎

patricksimonian:
personally I am okay with it actually. the role is limited. Only can post to devhub. If it is compromised I can delete the web hoook

PR #1085, opened 2019-11-05
add agile topic page into journey

w8896699:
Integrate topic into journey.
Only 'agile process' now,


app-web/src/components/Journey/JourneyMap.js
```diff
@@ -44,6 +44,7 @@ export const JourneyMap = ({ title, link, stops }) => {
             title
           )}
         </TopicTitle>
+        <div>{description}</div>
```
patricksimonian: missing props types for description. Also was the topic container using a `div` container? I don't think this is the best html tag to use.

PR #1086, opened 2019-11-05
Pipeline/remove deployments

patricksimonian:
## Summary
there is an issue with the github deployments api not being able to create deployments for m aster because of failing checks. This is odd since i've asked the api to ignore checks. to prevent this from being a blocker, the deployments api code has been skipped.


patricksimonian:
not merging yet. it appears we are having intermittent success with the scheduled build creating deployments

PR #1110, opened 2019-11-13
improve search function, search results that are local to the devhub …

w8896699:
Local resources will show up while other resources still loading.

### DIFF
--- a/app-web/src/components/Home/ResourcePreview.js
+++ b/app-web/src/components/Home/ResourcePreview.js
@@ -15,7 +15,7 @@ limitations under the License.
 
 Created by Patrick Simonian
 */
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
 import css from '@emotion/css';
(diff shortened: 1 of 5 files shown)


app-web/src/pages/index.js
```diff
@@ -407,6 +413,7 @@ export const Index = ({
         resultCount={totalSearchResults}
       />
       <Main>
+        {windowHasQuery && siphonResources}
```
w8896699: if search without query, the location.search will be ?q='' which will lead a bug. (cuz windowHasQuery is True)

app-web/src/components/Home/ResourcePreview.js
```diff
@@ -92,6 +92,14 @@ export const ResourcePreview = ({ title, link, resources, filters, amountToShow,
   let [activeFilters, setActiveFilters] = useState(['All']);
   const extraItemsToShow = 6; //two more row of card in the page after click 'see more'
 
+  useEffect(() => {
+    setResources(resources);
+
+    return () => {
+      setResources();
```
patricksimonian: setResources to `undefined`?

app-web/src/pages/index.js
```diff
@@ -251,7 +251,11 @@ export const Index = ({
     DYNAMIC_TOPIC_PATHS.featured,
     FEATURED_CONTENT,
   );
-
+  // dynamic sources all load at different times, this function returns false when all have completed loading
+  let [searchSourcesLoading, setLoading] = useState(searchGate.loading);
+  if (!!searchSourceResults.rocketchat) {
```
patricksimonian: what does this step do?

app-web/src/pages/index.js
```diff
@@ -407,7 +413,12 @@ export const Index = ({
         resultCount={totalSearchResults}
       />
       <Main>
-        {windowHasQuery && searchSourcesLoading ? <Loading message="loading" /> : content}
+        {windowHasQuery && queryParam.q !== '' && siphonResources}
```
patricksimonian: can you double check if we are already making this check in `queryisEmpty`

PR #1112, opened 2019-11-13
use right arrow instead of down arrow

w8896699:
Based on Shea's suggestion, use right arrow instead on down arrow


patricksimonian:
I think this needs to be rebased so we can get this in

PR #1113, opened 2019-11-13
Create nodejs.yml

patricksimonian:
add github actions ci


patricksimonian:
missing some environnment variables that is causing the build to fail :(

PR #1118, opened 2019-11-14
Fix search source icons flicker

w8896699:
quick fix for icon UI delay show up issue

### DIFF
--- a/app-web/src/components/Search/SearchSources.js
+++ b/app-web/src/components/Search/SearchSources.js
@@ -11,6 +11,10 @@ const StyledLink = styled(Link)`
   margin: 0 5px;
 `;
 
+const StyledDiv = styled.div`
+  margin: 0 5px;
+`;
+
 export const SearchSourcesContainer = styled.div`
   display: flex;
   padding: 6px;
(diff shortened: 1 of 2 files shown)


app-web/src/components/Search/SearchSources.js
```diff
@@ -22,45 +27,55 @@ export const TEST_IDS = {
...
-export const SearchSources = () => {
+export const SearchSources = ({ searchSourcesLoading }) => {
   const { authenticated } = useAuthenticated();
   const iconProps = {};
 
-  if (!authenticated) {
+  if (!authenticated || searchSourcesLoading) {
     iconProps.style = {
       opacity: 0.65,
-      margin: '0 2px',
```
patricksimonian: why delete the margin and padding?

app-web/src/components/Search/SearchSources.js
```diff
@@ -22,45 +27,55 @@ export const TEST_IDS = {
...
+  } else {
+    iconProps.style = {};
   }
 
   const scrollOffset = -125;
 
   return (
     <SearchSourcesContainer data-testid={TEST_IDS.container}>
-      {authenticated ? (
+      {authenticated || !searchSourcesLoading ? (
```
patricksimonian: so if the user is authenticated or if search sources have stoppped loading? Is this the intended effect

patricksimonian: &&

PR #1144, opened 2019-11-21
Code quality improvement

w8896699:
Fix a little bit code that code climate point out.

### DIFF
--- a/app-web/__tests__/gatsby/getFilesFromRegistry.test.js
+++ b/app-web/__tests__/gatsby/getFilesFromRegistry.test.js
@@ -1,4 +1,5 @@
 import { getFilesFromRegistry, expandRegistry } from '../../gatsby/utils/githubRaw';
+import { registry } from '../../plugins/gatsby-source-github-all/__tests__/registryHelpers';
 
 const SOURCE_1 = {
   sourceType: 'github',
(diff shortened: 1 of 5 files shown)


w8896699:
Index test doesn't pass...Can't really figure out why.

patricksimonian:
if you are exporting files as `fixtures` they should be in fixtures.

app-web/src/pages/index.js
```diff
@@ -206,11 +206,8 @@ export const Index = ({
   );
 
   let totalSearchResults = 0;
-
-  const resourcesNotFound =
-    !queryIsEmpty &&
-    (!results || (results.length === 0 && windowHasQuery)) &&
-    isEmpty(searchSourceResults);
+  const resultHasNothing = results && results.length === 0;
```
patricksimonian: `searchResultsDontExist`

patricksimonian: `noSearchResults` thats probably better..

app-web/__tests__/pages/Index.test.js
```diff
@@ -123,6 +123,7 @@ describe('Home Page', () => {
     queryString.parse.mockReturnValue({ q: 'foo' });
     // rerender stubbing no results for resources
     useSearch.mockReturnValue([]);
+    removeUnwantedResults.mockReturnValue([]);
```
w8896699: For some mystical reason, if removeUnwantedResults is imported from helper into page/index(or anywhere else),
it will return [] as undefined in the initial render. so I just mock the return value here to pass the test as in this case, useSearch is already empty array.

app-web/plugins/gatsby-source-github-all/__tests__/registryHelpers.js
```diff
@@ -37,7 +37,7 @@ const SOURCE_3 = {
   },
 };
 
-const registry = [
+export const registry = [
```
patricksimonian: let's not move __fixtures__ outside of that directory. This shoould be the other way around if you want to reduce duplication.

app-web/__tests__/gatsby/getFilesFromRegistry.test.js
```diff
@@ -27,52 +28,6 @@ const SOURCE_3 = {
   },
 };
 
-const registry = [
```
patricksimonian: lets move the registry into a fixture file if it already hasn't been done

w8896699: It has been moved to __fixtures__/fixtures.js

PR #1148, opened 2019-11-26
Add github sources' title as title with helmet

w8896699:
make our sites more discoverable by search engines.

### DIFF
--- a/app-web/src/templates/SourceGithub_default.js
+++ b/app-web/src/templates/SourceGithub_default.js
@@ -38,6 +38,7 @@ import {
   SidePanel,
   MarkdownBody,
 } from '../components/GithubTemplate/common';
+import { Helmet } from 'react-helmet';
 
 class SourceGithubMarkdownDefault extends React.Component {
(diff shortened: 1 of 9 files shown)


app-web/src/templates/SourceGithub_default.js
```diff
@@ -69,6 +70,9 @@ class SourceGithubMarkdownDefault extends React.Component {
 
     return (
       <Layout>
+        <Helmet>
```
patricksimonian: code is duplicated

app-web/src/templates/StandAlone_GitHubRaw.js
```diff
@@ -138,6 +139,9 @@ export const StandAloneGitHubRawResource = ({ data: { githubRaw } }) => {
   const [owner, repo] = html_url.replace('https://github.com/', '').split('/');
   return (
     <Layout>
+      <Helmet>
```
patricksimonian: code is duplicated

PR #1188, opened 2019-12-13
Integrate Cypress E2E tests into github actions

patricksimonian:
## Summary
Cypress now has a supported github action that we can leverage


.github/workflows/cypress.yml
```diff
@@ -0,0 +1,38 @@
...
+            ${{ runner.OS }}-build-${{ env.cache-name }}-
+            ${{ runner.OS }}-build-
+            ${{ runner.OS }}-
+      - uses: bahmutov/npm-install@v1
+        with:
+          working-directory: app-web
+      # Install NPM dependencies, cache them correctly
+      # and run all Cypress tests
+      - name: Cypress run
+        uses: patricksimonian/github-action@6dc0f8790dccace7167b12f9aaa0e28b7dc81c7c
```
ShellyXueHan: is this token okay to be public? consider github secret?

patricksimonian: thats a commit id :)

PR #1192, opened 2019-12-17
Imp/#1143 code clean up

w8896699:
we are now using algolia search, so clean up code that use for previous two search plugins.


.github/workflows/nodejs.yml
```diff
@@ -43,4 +43,4 @@ jobs:
         MATOMO_API_KEY: ${{secrets.MATOMO_API_KEY}}
         GATSBY_ALGOLIA_APP_ID: ${{secrets.GATSBY_ALGOLIA_APP_ID}}
         GATSBY_ALGOLIA_SEARCH_KEY: ${{secrets.GATSBY_ALGOLIA_SEARCH_KEY}}
-        GATSBY_ALGOLIA_INDEX_NAME_SUFFIX: ${{secrets.GATSBY_ALGOLIA_INDEX_NAME_SUFFIX}}
+        NODE_env: ${{secrets.NODE_env}}
```
patricksimonian: NODE_ENV

patricksimonian: and it does'nt need to be secret. this can just be 'test'

app-web/gatsby-config.js
```diff
@@ -22,7 +22,7 @@ const eventbritePlugin = () =>
     : undefined;
 
 const algoliaPlugin = () =>
-  process.env.GATSBY_ALGOLIA_INDEX_NAME_SUFFIX !== 'test'
+  process.env.NODE_env !== 'test'
```
patricksimonian: NODE_ENV

PR #1218, opened 2020-01-13
Upgrade NPM packages via npm audit fix

patricksimonian:
## Summary
Routine npm package upgrade :)

## Other Changes
- `@octokit/rest` was failing upon import of the library during gatsby builds, this was replaced with a function to fetch github contents. This fixes that issue and removes @octokit as a dependancy


patricksimonian:
> Are you able to build using npm run build ? I tried a npm run build on this branch using my machine and was getting the same error we were getting yesterday.

did you try running `npm i` before hand?

PR #1230, opened 2020-01-15
Algolia/1227 pipeline

patricksimonian:
## Summary

Add the remaining devops goodness for algolia search integration

TODO

- [x] update caddy deployment config env variable with algolia index suffix
- [x] add index promotion scripts to jenkins pipeline
- [x] normalize algolia indices


patricksimonian:
as a note I'm okay with e2e tests failing right now. This should be the case since they are made against elastic lunr and not algolia. We will convert the e2e test in the integration branch #1157

Issue #1245, opened 2020-01-27
Filtering on homepage does not reset on a new search


patricksimonian:
**Describe the bug**
When searching you can filter the results by their resource type. if you do another search after clicking the filters they do not reset

**To Reproduce**
1. Go to 'home page'
2. search openshift
3. click the documentation pill
4. search agile
5. notice pills do not reset

**Expected behavior**
Pills should reset everytime a search has been done

** Additional Context **

IMPO i think we should be passing the filter parameters into the url to maintain state so that filters a bookmarkable. It should also make it easy to fix this bug since the url would be the single source of truth.

PR #1246, opened 2020-01-28
Fix and issue where page refreshes ruin the CSS

### LINKED ISSUE #1149: The main page css is not correct
Screen Shot 2019-11-27 at 3 23 38 PM
The css comonent seems not correct in dev after the first search, after hard referesh, it's gone

patricksimonian:
## Summary
I think this was a case of the index page doing too much. It looked like it did multiple re-renders which may bork up something internally. 
Fixes #1149 
## Changes

Completely refactored the homepage to be smaller and more concise. In addition, I added refactored using TDD with Cypress e2e tests.

## Additional Notes

During my refactor I have noticed  a  bug with the filters which has been noted #1245


jas29:
> Deployment in Dev still shows this issue
> image

Sorry looking at an outdated deployment

PR #1276, opened 2020-02-11
Fix filters not resetting on new searches

### LINKED ISSUE #1245: Filtering on homepage does not reset on a new search

patricksimonian:
## Summary
Fixes #1245 
## Notable Changes 
- updated E2E test and verified works

### DIFF
--- a/app-web/cypress/integration/Home/search.spec.js
+++ b/app-web/cypress/integration/Home/search.spec.js
@@ -84,7 +84,7 @@ describe('Searching from homepage', () => {
     cy.contains(/No resources found :\(/i);
   });
 
-  it.skip('can toggle search filters correctly', () => {
+  it('can toggle search filters correctly', () => {
     cy.log('entering openshift as search since it gives a lot of results');
     cy.visit('?q=openshift');
     cy.log('Result count pill should be visible');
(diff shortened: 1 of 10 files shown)


app-web/src/components/Home/ResourcePreview.js
```diff
@@ -84,49 +88,92 @@ export const TEST_IDS = {
...
+    const searchMap = {
+      ...queryParam,
+      [FILTER_QUERY_PARAM]: toggledFilters.length === filtersList.length ? [] : toggledFilters,
+    };
+    const searchString = queryString.stringify(searchMap);
+
     setCount(amountToShow);
+    // replacing state so that back button doesn't get poluted with filter changes
+    // also this will trigger the url to change but will not cause a page refresh unlike gatsby.navigate
+    window.history.replaceState({}, 'foo', `${pathname}?${searchString}`);
```
jas29: Used window.replacestate instead of gatsby navigate due to auto scolling in the naviagte function

app-web/cypress/integration/Home/search.spec.js
```diff
@@ -113,6 +116,32 @@ describe('Searching from homepage', () => {
...
+
+    cy.get('@firstFilterPill')
+      .should('have.attr', 'data-active')
+      .and('eq', 'false');
+
+    cy.log('The All pill should be automatically toggled when all the pills are selected');
+    cy.visit('?q=openshift');
+
+    cy.get('[data-testclass="resource-preview-pill"]')
+      .as('allPills')
```
patricksimonian: no need for alias

PR #1297, opened 2020-02-29
Include journeys and topics as cards in search results

jas29:
Fixes issue #1283


app-web/src/components/Card/Card.js
```diff
@@ -76,6 +76,7 @@ export const Card = ({
   link,
   renderBody,
   renderHeader,
+  slug,
```
patricksimonian: not needed, this would be captured in the ...rest

app-web/src/components/Card/CardsInColumns.js
```diff
@@ -35,7 +35,7 @@ const CardsInColumns = ({ cards }) => {
         title={card.fields.title}
         description={card.fields.description}
         image={card.fields.image}
-        link={card.fields.standAlonePath}
+        link={card.fields.standAlonePath || getFirstNonExternalResource(card.connectsWith)}
```
patricksimonian: we  want to move this implementation outside of the comopnent that calls  it. This is better places  at the index.js level

app-web/src/pages/index.js
```diff
@@ -61,16 +67,29 @@ export const Index = ({
     allEventbriteEvents.edges,
   ]);
 
+  const journeydata = useMemo(() => flattenGatsbyGraphQL(allJourneyRegistryJson.edges), [
+    allJourneyRegistryJson.edges,
+  ]);
+
+  const topicdata = useMemo(() => flattenGatsbyGraphQL(allTopicRegistryJson.edges), [
```
patricksimonian: this is where you want to map the topics data  to have  a standAlonePath

app-web/src/templates/resourceType.js
```diff
@@ -211,5 +213,38 @@ export const ResourceTypeQuery = graphql`
         }
       }
     }
+    allJourneyRegistryJson {
```
patricksimonian: why is this needed here?

PR #1299, opened 2020-03-04
events packaged more neatly

patricksimonian:
## Summary
Update Events page to reduce its build size  - improving  site performance.


patricksimonian:
@jas29 maybe we can get this merged shotrly after #1347

jas29:
> @jas29 maybe we can get this merged shotrly after #1347

yes will do that, once 1347 is deployed and merged

Issue #1293, opened 2020-02-28
Clicking "persona" pills doesn't show all resources associated with a given persona


sheaphillips:
**Describe the bug**
Clicking on persona "pill" on a standalone page *should* go to a page showing all resources associated with the clicked persona. 

**To Reproduce**
1. Go to a standalone page that has persona metadata. (e.g. https://developer.gov.bc.ca/Alignment)
2. In the "Resource Information" section at the top of the page, click on the pill labelled 'Developer'.  You will be taken to a search results page. 
3. The page will be as shown in attachment (a single resource shown).

**Expected behavior**
A set of search results including the many resources with "Developer" metadata persona value.

**Screenshots**
<img width="1215" alt="Screenshot 2020-02-27 16 49 28" src="https://user-images.githubusercontent.com/698658/75500485-ed977200-5981-11ea-9957-6ed4bf380d08.png">

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Safari, Chrome
 - Version 13.0.5, 80.0.3987.122

**Additional context**
The example provided is for the "Developer" pill.  The Product Owner and Designer pills seem to show an empty list.

patricksimonian:
hmm this was definitely missed when doing the algolia integration. I will look into how to get filters applied again. It should be easy enough!

PR #1315, opened 2020-03-12
Persona Pill Resources

jas29:
Solves issue #1293

### DIFF
--- a/app-web/src/utils/algolia.js
+++ b/app-web/src/utils/algolia.js
@@ -11,6 +11,9 @@ const algoliaIndexQuery = `{
             title
             tags
           }
+          frontmatter {
+            personas
+          }
           content: excerpt(pruneLength: 1000)
         }
         internal {
@@ -62,6 +65,7 @@ const algoliaIndexQuery = `{
           title
           description
           tags
+          personas
         }
         internal {
           type
@@ -101,19 +105,23 @@ export const reduceNodesForIndex = nodes => {
     } = node;
     let fields,
       content,
+      persona,
       id = node.id;
 
     if (type === 'GithubRaw') {
       fields = node.childMarkdownRemark.fields;
       content = node.childMarkdownRemark.content;
+      persona = node.childMarkdownRemark.frontmatter.personas;
     } else {
       content = '';
       fields = node.fields;
+      persona = node.fields.personas;
     }
     return {
       fields,
       content,
       id,
+      persona,
       [NODE_TYPE_FIELD_NAME]: type,
     };
   });
--- a/app-web/src/utils/hooks.js
+++ b/app-web/src/utils/hooks.js
@@ -51,27 +51,20 @@ export const useSearch = query => {
   const [results, setResults] = useState([]);
   const index = searchClient.initIndex(`Devhub-Algolia-${ALGOLIA_INDEX_SUFFIX}`);
   useEffect(() => {
+    let params;
+    let options = {};
     if (query && query.includes('persona:')) {
-      index
-        .search('', {
-          facetFilters: [query],
-        })
-        .then(res => {
-          setResults(res.hits);
-        });
-    } else if (query) {
-      // if no query, I.E query is '', algolia will return all index....
-      index
-        .search({
-          query: query,
-          hitsPerPage: 200, //set it to a large number, the default value is 200, which some search result is over that number
-        })
-        .then(res => {
-          setResults(res.hits);
-        });
+      params = '';
+      options.facetFilters = [query];
     } else {
-      setResults([]);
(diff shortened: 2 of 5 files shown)


app-web/src/utils/algolia.js
```diff
@@ -101,19 +105,23 @@ export const reduceNodesForIndex = nodes => {
     } = node;
     let fields,
       content,
+      persona,
       id = node.id;
 
     if (type === 'GithubRaw') {
       fields = node.childMarkdownRemark.fields;
       content = node.childMarkdownRemark.content;
+      persona = node.childMarkdownRemark.frontmatter.personas;
```
patricksimonian: isn't personas a field

app-web/src/utils/hooks.js
```diff
@@ -52,15 +52,27 @@ export const useSearch = query => {
...
-          query: query,
-          hitsPerPage: 200, //set it to a large number, the default value is 200, which some search result is over that number
-        })
-        .then(res => {
-          setResults(res.hits);
-        });
+      if (query.includes('personas:')) {
+        let personaName = query.split(':')[1];
+        const filters = 'persona :' + personaName;
+        index
```
patricksimonian: those two querieis could be refacgtored. only thing that changes is the search object

sheaphillips:
Any chance you could expand scope on this item to deal with labels also? There’s a desperate ticket for it, but I think it will be at least as useful as fixing personas...

app-web/src/utils/algolia.js
jas29: it is a field for Devhub Siphon Nodes, however it exists within the frontmatter object for github raw nodes

jas29:
> Any chance you could expand scope on this item to deal with labels also? There’s a desperate ticket for it, but I think it will be at least as useful as fixing personas...
> […](#)
> Sent from my iPhone
> On Mar 20, 2020, at 9:06 AM, Patrick Simonian ***@***.***> wrote: ﻿ @patricksimonian requested changes on this pull request. oops didn't mean to approve — You are receiving this because your review was requested. Reply to this email directly, view it on GitHub, or unsubscribe.

Yup Ill look into it :)

patricksimonian:
@jas29 I'd do this in another pr. that work doesn't have to do with search so much as it has to do with creating dynamic topics based on that label. I think its an easy todo with a few challenges on labels. We shouldn't fire this off immediatly. Stay the course :)

sheaphillips:
It's cool to do it as a separate PR.  I don't think we need dynamic topics,
though.  I *think* it's the same as the persona stuff - index the labels,
and then show clickable labels on pages/cards that run searches. (eg.
.../?query=label:cloud&label=azure).

app-web/src/utils/algolia.js
patricksimonian: lets make this a field for github raw nodes.

patricksimonian: can be done in another pr

app-web/src/utils/hooks.js
```diff
@@ -51,7 +51,15 @@ export const useSearch = query => {
   const [results, setResults] = useState([]);
   const index = searchClient.initIndex(`Devhub-Algolia-${ALGOLIA_INDEX_SUFFIX}`);
   useEffect(() => {
-    if (query) {
+    if (query && query.includes('persona:')) {
+      index
```
patricksimonian: the search operation is the same. I think the code would look cleaner if we just did the search once and modified the arguments inside of it. 

```js

if foo
query = '';
options = {
  facetFilters: bar
}
else 
query = blah
options = {...}

// do search here

File: app-web/src/utils/hooks.js [hunk 4]
```diff
@@ -51,19 +51,20 @@ export const useSearch = query => {
...
-          setResults(res.hits);
-        });
+    let params;
+    let options = {};
+    if (query && query.includes('persona:')) {
+      params = '';
+      options.facetFilters = [query];
     } else {
-      setResults([]);
+      params = null;
```
patricksimonian: seems like a bug?

app-web/src/utils/hooks.js
```diff
@@ -51,16 +51,17 @@ export const useSearch = query => {
...
-          hitsPerPage: 200, //set it to a large number, the default value is 200, which some search result is over that number
-        })
-        .then(res => {
-          setResults(res.hits);
-        });
+      if (query && query.includes('persona:')) {
+        options.facetFilters = [query];
+      } else {
+        options.query = query;
+        options.hitsPerPage = 200;
```
patricksimonian: 200 is the default hitsPerPage amount

PR #1361, opened 2020-05-04
Enable Cors on Devhub-API

jas29:
## Summary
Enable cors between devhub and our microservice and generate pull request using the github api

### DIFF
--- a/api/src/index.js
+++ b/api/src/index.js
@@ -30,6 +30,13 @@ const app = express();
 // middlewares
 app.use(bodyParser.json());
 
+// Add access control header (cors)
+app.use(function(req,res,next){
+   res.header("Access-Control-Allow-Origin", "http://localhost:8000/contentContribution")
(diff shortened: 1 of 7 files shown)


api/src/index.js
```diff
@@ -30,6 +31,22 @@ const app = express();
 // middlewares
 app.use(bodyParser.json());
 
+
+var corsOptions = {
+   origin: `${process.env.CORS_URL}/contentContribution`,
```
patricksimonian, line 36: avoid paths. /contentContribution is too specific

api/src/index.js
```diff
@@ -30,6 +31,22 @@ const app = express();
...
+   optionsSuccessStatus: 200
+}
+
+app.use(cors(corsOptions),function(req,res){
```
patricksimonian, line 40: just need app.use(cors(...))

PR #1362, opened 2020-05-04
Add create topics api to the Devhub API

patricksimonian:
## Summary
TODO

PR #1373, opened 2020-05-21
Feat/topic integ

jas29:
## Summary
Quick Summary Here
Topic Integration PR on top of #1362


api/src/controllers/topics.js
```diff
@@ -0,0 +1,108 @@
...
+};
+
+export const createFile = async (owner, repo) => {
+  const path = 'app-web/topicRegistry/topicName.json';
+  const message = 'test commit';
+  const branch = 'refs/heads/test';
+  const content = Base64.encode(
+    JSON.stringify({ name: 'topicName', description: 'lorem ipsum' }, null, 2),
+  );
+  const committer = { email: 'karansaini29@gmail.com', name: 'jas29' };
```
patricksimonian, line 51: no hard coding, add the variables or add nothing.

api/src/controllers/topics.js
```diff
@@ -0,0 +1,108 @@
...
+      // validate topic doesn't already exist
+      // validate topic sources are valid
+      await createNewRefFromBase(owner, repo, ref);
+      // create a new file with contents
+      await createFile(owner, repo, bodyData, ref, topicName, email, name);
+      // commit  to branch
+      // make pr against ref to base using templates
+      await createPullRequest(owner, repo, defaultBranch, topicName, ref);
+    } catch (e) {
+      console.log(e);
```
patricksimonian, line 97: should send an error, this will just send no response if theres a failure.

api/src/utils/github.js
```diff
@@ -0,0 +1,14 @@
+import octokit from '../octokit';
+import { PULL_REQUEST_STATE } from '../constants';
+
```
patricksimonian, line 3: this is where you octokit wrappers should live

api/src/controllers/topics.js
```diff
@@ -0,0 +1,108 @@
...
+    }
+  }
+  // check if an open pr exists
+
+  // create a git branch on the remote with a good naming convention
+  // something like contribution/username-[known-identifier]
+
+  // create the topic and place in branch
+  // create the pr
+  res.send('ok');
```
patricksimonian, line 107: should send an actual payload instead of just ok

api/src/controllers/topics.js
```diff
@@ -0,0 +1,108 @@
...
+  const ref = `refs/heads/createTopic/${topicName}`;
+  console.log(bodyData);
+  // console.log('DEFAULT branch', defaultBranch, req.body);
+  if (await openPullExistsForBranch(branchName, repo, owner)) {
+    // add payload to pull request as new commit
+  } else {
+    // create branch
+    try {
+      // validate topic schema
+      validate(req.body.topic);
```
patricksimonian, line 87: does this validate fn return a  boolean? I wonder i f we need to check that out.

api/src/config/index.json
```diff
@@ -1,4 +1,13 @@
 {  
+  "github": {
+    "repo": "devhub-app-web",
+    "owner": "jas29",
+    "email": "karansaini29@gmail.com",
```
patricksimonian, line 5: are you sure you want this to be in the commit history?

api/src/controllers/topics.js
```diff
@@ -0,0 +1,57 @@
...
+        // TODO -- validate topic sources are valid
+        // create a git branch on the remote with a naming convention [createTopic/topicname]
+        await createNewRefFromBase(owner, repo, ref);
+        // // create a new file with contents
+        await createFile(owner, repo, bodyData, ref, topicName, email, name);
+        // commit  to branch
+        // make pr against ref to base using templates
+        const pullRequest = await createPullRequest(owner, repo, defaultBranch, topicName, ref);
+        // URL to the pull Request created ..
+        const pullRequest_url = pullRequest.data.html_url;
```
patricksimonian, line 45: pullRequestUrl. JS is typically in camelCase

api/src/controllers/topics.js
```diff
@@ -0,0 +1,57 @@
...
+      } else {
+        status = '400';
+        statusMessage = 'Bad Request';
+      }
+    }
+  } catch (e) {
+    status = e.status;
+    statusMessage = e;
+  }
+  res.status(status).send(statusMessage);
```
patricksimonian, line 56: this may work for now but because we are sending json data forward, it is  pretty common to get JSON back. The JSON payload has the info in it. This way it is also organized. m

`res.json({prUrl: ...})`;

its better to simplify then make oppinions on the backend. If you send the PR URL the front end can represent it in anyway it wants. 

What you do want to do is at good console.logs so that the pods stdout has an audit of PR's being created.

api/src/controllers/topics.js
```diff
@@ -0,0 +1,62 @@
...
+        // create a git branch on the remote with a naming convention [createTopic/topicname]
+        await createNewRefFromBase(owner, repo, ref);
+        // // create a new file with contents
+        await createFile(owner, repo, bodyData, ref, topicName, email, name);
+        // commit  to branch
+        // make pr against ref to base using templates
+        const pullRequest = await createPullRequest(owner, repo, defaultBranch, topicName, ref);
+        // URL to the pull Request created ..
+        const pullRequestUrl = pullRequest.data.html_url;
+        response.statusMessage = pullRequestUrl;
```
patricksimonian, line 51: this is not a really great response payload. Lets take a think about it from the front perspective. 

We aren't charged for how many  key value pairs we introduce into a json payload. 

why not 
```js
{
  prUrl: ...,
  message: 'if needed',
}
```

you are using a key with two conceptually different values.

api/src/controllers/topics.js
```diff
@@ -48,7 +48,7 @@ export const createOrUpdateTopic = async (req, res) => {
         const pullRequest = await createPullRequest(owner, repo, defaultBranch, topicName, ref);
         // URL to the pull Request created ..
         const pullRequestUrl = pullRequest.data.html_url;
-        response.statusMessage = pullRequestUrl;
+        response.prUrl = pullRequestUrl;
```
patricksimonian, line 51: console.log out  that a pr was created

- name of the topic,
- the pr url

app-web/src/components/Form/TopicForm.js
```diff
@@ -1,17 +1,63 @@
...
+import Loading from '../UI/Loading/Loading';
+import { useKeycloak } from '@react-keycloak/web';
+import { DEVHUB_API_URL } from '../../constants/api';
 
 export const TopicForm = () => {
+  const [loading, setLoading] = useState(false);
+
+  const [response, setResponse] = useState({ status: '', data: '' });
+
+  const [showResponseMessage, setResponseMessage] = useState(false);
```
patricksimonian, line 23: whats wrong with this 😏 

> hint i've explained this to you 5 times now

PR #1402, opened 2020-06-23
Create Edit a topic form for the devhub.

jas29:
Add edit a topic func. to the devhub.

Have an edit icon on next to each topic which re-directs them to the edit page.
Have the form pre-populated with the current topic data.


app-web/src/components/Form/ExistingResourceForm.js
```diff
@@ -1,93 +1,199 @@
...
+import StyledButton from '../UI/Button/Button';
+import { Form } from 'react-final-form';
+import arrayMutators from 'final-form-arrays';
+import { FieldArray } from 'react-final-form-arrays';
 
 export const ExistingResourceForm = () => {
-  const [currentFormValue, setFormValue] = useState('');
-  let content;
+  const topicName = location.pathname.replace('/editTopic/', '').replace(/-/g, ' ');
+  let topicData = dataQuery(topicName);
```
patricksimonian, line 18: whats data query?

app-web/src/components/Form/ExistingResourceForm.js
```diff
@@ -1,93 +1,199 @@
...
+} from './form';
+import StyledButton from '../UI/Button/Button';
+import { Form } from 'react-final-form';
+import arrayMutators from 'final-form-arrays';
+import { FieldArray } from 'react-final-form-arrays';
 
 export const ExistingResourceForm = () => {
-  const [currentFormValue, setFormValue] = useState('');
-  let content;
+  const topicName = location.pathname.replace('/editTopic/', '').replace(/-/g, ' ');
```
patricksimonian, line 17: theres better ways of doing this. Paths are also carried in the `location` prop when using client side only routes. 

So you could do something like `<EditTopic path="/editTopic/:name" />`

and reference it within the `location` prop

app-web/src/components/Topic/index.js
```diff
@@ -79,6 +82,9 @@ const Topic = ({ title, description, link, resources, ...rest }) => (
   <Container {...rest}>
     <TopicDecorativeBar color={TOPICS} />
     <PreviewContainer>
+      <Link to={`/editTopic/${slugify(title)}`}>
```
patricksimonian, line 85: can you double check to see if topics odn't already have some type of data field that has a slugified page path

app-web/src/pages/editTopic.js
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <Layout>
+      <SEO title="Edit topic" />
+      <Main>
+        <Title
+          title="Content Contribution"
+          subtitle={!isAuthenticated ? 'Login via IDIR or Github to continue' : 'Edit a topic'}
+        />
+        {isAuthenticated ? (
+          <ExistingResourceForm></ExistingResourceForm>
```
patricksimonian, line 24: whats wrong with this?

app-web/src/pages/editTopic.js
```diff
@@ -0,0 +1,33 @@
...
+        ) : (
+          <h4 css={{ color: 'red' }}>Not Authorized</h4>
```
patricksimonian, line 26: I wonder if it would be best to just redirect them to a 401 page over just saying Not authorized in 'murder red'

app-web/src/components/Form/ExistingResourceForm.js
jas29: dataQuery is the functions that grabs the topic data from via a graphql query

patricksimonian: okay so why not call it `getTopicDataByName` :)

patricksimonian: maybe think about not what the thing is. but what does it do? Name accordingly

app-web/src/components/Form/TopicForm.js
```diff
@@ -7,27 +7,66 @@ import {
...
+  // Configuration for edit a topic
+
+  if (operation === 'edit') {
+    const topicName = location.pathname
+      .split('/')
+      .pop()
+      .replace(/-/g, ' ');
+    const topicData = dataQuery(topicName);
+    config = {
+      apiEndPoint: '/v1/edit/topic/',
```
patricksimonian, line 48: this will probably change but good enough for now.

app-web/src/components/Form/form.js
```diff
@@ -55,6 +55,8 @@ export const StylesWrapper = styled.div`
   input {
     display: block;
     font-size: 18px;
+    width: 500px;
+    height: 50px;
```
patricksimonian, line 59: this feels wrong, Why is height set? usually height is the sum of spacing and font size for input fields.

app-web/src/components/Topic/index.js
patricksimonian: was there no slug??

jas29: nope it's not there

app-web/src/components/Form/TopicForm.js
```diff
@@ -90,6 +121,12 @@ export const TopicForm = () => {
               {({ fields }) =>
                 fields.map((name, index) => (
                   <Fragment key={`${name}~${index}`}>
+                    <RemoveButton
+                      type="button"
+                      onClick={() => values.sources.length > 1 && fields.remove(index)}
+                    >
+                    ❌
```
patricksimonian, line 128: can we replace this with a font awesome icon?

app-web/src/components/Form/TopicForm.js
```diff
@@ -90,6 +121,12 @@ export const TopicForm = () => {
               {({ fields }) =>
                 fields.map((name, index) => (
                   <Fragment key={`${name}~${index}`}>
+                    <RemoveButton
+                      type="button"
+                      onClick={() => values.sources.length > 1 && fields.remove(index)}
```
patricksimonian, line 126: if field .sources is === to 1 then does this button show but not work?

app-web/src/components/Form/TopicForm.js
```diff
@@ -140,31 +177,34 @@ export const TopicForm = () => {
...
-  properties.url = source.url;
+  properties.url = source.sourceProperties.url;
   if (source.sourceType === 'github') {
-    properties.owner = source.owner;
-    properties.repo = source.repo;
-    properties.files = [source.file];
+    properties.owner = source.sourceProperties.owner;
+    properties.repo = source.sourceProperties.repo;
+    properties.files =
+      operation === 'create'
```
patricksimonian, line 202: can you explain whats going on here?

app-web/src/components/Form/TopicForm.js
```diff
@@ -140,31 +177,34 @@ export const TopicForm = () => {
...
+        ? [source.sourceProperties.files.split(',')]
+        : String(source.sourceProperties.files).split(',');
```
patricksimonian, line 204: This seems funny... You typically do not use the String constructor without `new`

app-web/src/components/Form/TopicForm.js
```diff
@@ -173,30 +213,70 @@ const SubForm = (sourceType, name) => {
...
+            }
+            resourceType
+          }
+        }
+      }
+    }
+  `);
+  const topicsNode = useMemo(() => flattenGatsbyGraphQL(topics.edges), [topics.edges]);
+  let topicData = {};
+  topicsNode.map(topic => {
```
patricksimonian, line 274: Incorrect use of map. You are looking for `.find` or `.findIndex`

app-web/src/components/Form/form.js
```diff
@@ -33,6 +33,13 @@ export const StyledLabel = styled.label`
   display: block;
 `;
 
+export const RemoveButton = styled.button`
+  border: none;
+  position: absolute;
+  background: white;
+  float: right;
```
patricksimonian, line 40: floats haven't been used since 1992 :)

app-web/src/components/Form/TopicForm.js
jas29: There is a slight change only with the **files** input field. When creating a topic, the input field works as usual, however when editing a topic, the files  data comes in as an array. So there needs to be converted to a string before any other operations can be done.

app-web/src/utils/contentContribution.js
```diff
@@ -0,0 +1,43 @@
...
+ * @param {Object} source the source (could be web or github)
+ * @param {String} source.sourceType
+ * @param {Object} source.sourceProperties
+ * @param {String} operation the form mode (edit or create)
+ */
+export const reduceFormSourceToSourceProperties = ({ sourceType, sourceProperties }) => {
+  if (sourceType === 'github') {
+    return {
+      ...sourceProperties,
+      files: sourceProperties.files.split(',').map(file => file.trim()),
```
jas29, line 13: This will still not work for the edit topic feat. The file data is coming in as an array. You can't perform .trim and .split on an array. Needs to be converted to a string first.

patricksimonian: im still really confused on how the implementations change between edit and create

app-web/gatsby/onCreatePage.js
```diff
@@ -27,6 +27,11 @@ module.exports = async ({ page, actions }) => {
     page.matchPath = `/topic/*`;
     createPage(page);
   }
-  // Delete the content contribution page so that it does not show up on Devhub.
+
+  // client only pages for edit topic functionality.
```
patricksimonian, line 31: the editTopic page will be staic now?

PR #1414, opened 2020-07-03
Edit a topic functionality for devhub api

jas29:
Enable the Devhub API, to edit topics and make a pull request.

### DIFF
--- a/api/src/constants.js
+++ b/api/src/constants.js
@@ -7,3 +7,20 @@ export const RESOURCE_VERBS = {
   UPDATE: 'UPDATE',
   CREATE: 'CREATE',
 };
+
+export const FILE_PATH = 'app-web/topicRegistry/';
+
+export const COMMIT_MESSAGE = {
+  CREATE: 'add new topic',
+  UPDATE: 'edit topic',
+};
+
+export const PR_TITLE = {
+  CREATE: 'Suggest new topic to the devhub',
+  UPDATE: 'Suggest an edit to topic',
+};
+
+export const PR_BODY = {
+  CREATE: 'Add new topic to the devhub',
+  UPDATE: 'Edit topic to include new changes',
+};
--- a/api/src/constants.js
+++ b/api/src/constants.js
@@ -15,12 +15,14 @@ export const COMMIT_MESSAGE = {
   UPDATE: 'edit topic',
 };
 
-export const PR_TITLE = {
-  CREATE: 'Suggest new topic to the devhub',
-  UPDATE: 'Suggest an edit to topic',
-};
-
-export const PR_BODY = {
-  CREATE: 'Suggest a new topic',
-  UPDATE: 'Suggest new changes to topic',
+export const PR_BODY = (operation, topicName, topicDescription) => {
+  let prData = { title: '', body: '' };
+  if (operation === 'create') {
+    prData.title = `Suggest new topic to the devhub ${topicName}`;
+    prData.body = `Suggest a new topic ${topicName} \n ${topicDescription}`;
+  } else if (operation === 'update') {
+    prData.title = `Suggest an edit to topic ${topicName}`;
+    prData.body = `Suggest new changes to topic ${topicName} \n ${topicDescription}`;
+  }
+  return prData;
 };
--- a/api/src/constants.js
+++ b/api/src/constants.js
@@ -21,6 +21,6 @@ export const PR_TITLE = {
 };
 
 export const PR_BODY = {
-  CREATE: 'Add new topic to the devhub',
-  UPDATE: 'Edit topic to include new changes',
+  CREATE: 'Suggest a new topic',
+  UPDATE: 'Suggest new changes to topic',
 };
--- a/api/src/controllers/createTopic.js
+++ b/api/src/controllers/createTopic.js
@@ -1,0 +1,26 @@
+import { github } from '../config/index.json';
+import slugify from 'slugify';
+import dotenv from 'dotenv';
+import { githubHandler } from '../utils/githubhandler';
+
+dotenv.config();
+
+export const createTopic = async (req, res) => {
+  const operation = 'create';
+  const { repo, owner, defaultBranch } = github;
+  const bodyData = JSON.stringify(req.body, null, 2);
+  const topicName = slugify(req.body.name.toLowerCase(), '-');
+  const ref = `refs/heads/createTopic/${topicName}`;
+
+  const response = await githubHandler(
+    operation,
+    defaultBranch,
+    repo,
+    owner,
+    req,
+    ref,
+    bodyData,
+    topicName,
+  );
+  res.status(response.status).json(response);
+};
--- a/api/src/controllers/createTopic.js
+++ b/api/src/controllers/createTopic.js
@@ -1,9 +1,6 @@
 import { github } from '../config/index.json';
 import slugify from 'slugify';
-import dotenv from 'dotenv';
-import { githubHandler } from '../utils/githubhandler';
-
-dotenv.config();
+import { createPullRequestFromData } from '../utils/githubhandler';
 
 export const createTopic = async (req, res) => {
   const operation = 'create';
@@ -12,7 +9,7 @@ export const createTopic = async (req, res) => {
   const topicName = slugify(req.body.name.toLowerCase(), '-');
   const ref = `refs/heads/createTopic/${topicName}`;
 
-  const response = await githubHandler(
+  const response = await createPullRequestFromData(
     operation,
     defaultBranch,
     repo,
--- a/api/src/controllers/createTopic.js
+++ b/api/src/controllers/createTopic.js
@@ -5,7 +5,7 @@ import { createPullRequestFromData } from '../utils/githubhandler';
 export const createTopic = async (req, res) => {
   const operation = 'create';
   const { repo, owner, defaultBranch } = github;
-  const bodyData = JSON.stringify(req.body, null, 2);
+  const bodyData = req.body;
   const topicName = slugify(req.body.name.toLowerCase(), '-');
   const ref = `refs/heads/createTopic/${topicName}`;
 
@@ -14,7 +14,6 @@ export const createTopic = async (req, res) => {
     defaultBranch,
     repo,
     owner,
-    req,
     ref,
     bodyData,
     topicName,
--- a/api/src/controllers/editTopic.js
+++ b/api/src/controllers/editTopic.js
@@ -1,0 +1,22 @@
+import slugify from 'slugify';
+import { github } from '../config/index.json';
+import { githubHandler } from '../utils/githubhandler';
+
+export const editTopics = async (req, res) => {
+  const operation = 'update';
+  const { repo, owner, defaultBranch } = github;
+  const bodyData = JSON.stringify(req.body, null, 2);
+  const topicName = slugify(req.body.name.toLowerCase(), '-');
+  const ref = `refs/heads/editTopic/${topicName}`;
+  const response = await githubHandler(
+    operation,
+    defaultBranch,
+    repo,
+    owner,
+    req,
+    ref,
+    bodyData,
+    topicName,
+  );
+  res.status(response.status).json(response);
+};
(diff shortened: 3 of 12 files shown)


api/src/controllers/topics.js
```diff
@@ -18,14 +18,19 @@ const ajv = new Ajv();
...
   const bodyData = JSON.stringify(req.body, null, 2);
   const topicName = slugify(req.body.name.toLowerCase(), '-');
-  const ref = `refs/heads/createTopic/${topicName}`;
+  const ref =
+    operation === 'create'
+      ? `refs/heads/createTopic/${topicName}`
+      : `refs/heads/editTopic/${topicName}`;
   const response = await githubFunc(
+    operation,
     branchName,
```
patricksimonian, line 34: replace branchname with ref?

api/src/controllers/topics.js
```diff
@@ -41,6 +46,7 @@ export const createOrUpdateTopic = async (req, res) => {
 };
 
 const githubFunc = async (
```
patricksimonian, line 48: what does this do?

no js docs?

why is this in the same file if its a helper

api/src/controllers/topics.js
```diff
@@ -54,28 +60,35 @@ const githubFunc = async (
...
-        const pullRequest = await createPullRequest(owner, repo, defaultBranch, topicName, ref);
-        // URL to the pull Request created ..
-        const pullRequestUrl = pullRequest.data.html_url;
-        response.prUrl = pullRequestUrl;
-        //eslint-disable-next-line
+    // create new topic and branch
+    // validate topic schema
+    const isValidData = validate(req.body);
+    if (isValidData) {
+      // TODO -- validate topic doesn't already exist
```
patricksimonian, line 67: just make sure to create github issues for these TODO's so they are not missed. be descriptive. 

File name etc

api/src/routers/topics.js
```diff
@@ -5,4 +5,6 @@ const router = express.Router();
 
 router.post('/', createOrUpdateTopic);
 
+router.post('/edit/', createOrUpdateTopic);
```
patricksimonian, line 8: split up this function into create or update. Also technically this should be a router.put function

api/src/utils/github.js
```diff
@@ -44,28 +44,60 @@ export const createNewRefFromBase = async (owner, repo, ref) => {
...
+    });
+    return createFile;
+  }
 };
 
-export const createPullRequest = async (owner, repo, base, topicName, ref) => {
-  const title = `Add new topic ${topicName}`;
-  const body = `Add a new topic to the devhub named ${topicName}`;
+export const createPullRequest = async (operation, owner, repo, base, topicName, ref) => {
+  const title = operation === 'create' ? `Add new topic ${topicName}` : `Edit topic ${topicName}`;
```
patricksimonian, line 98: abstract content-y things into template functions or something similar. For example

```js
export const editTopicTitle = name => `...`;

File: api/src/constants.js [hunk 6]
```diff
@@ -7,3 +7,20 @@ export const RESOURCE_VERBS = {
...
+  UPDATE: 'edit topic',
+};
+
+export const PR_TITLE = {
+  CREATE: 'Suggest new topic to the devhub',
+  UPDATE: 'Suggest an edit to topic',
+};
+
+export const PR_BODY = {
+  CREATE: 'Add new topic to the devhub',
```
patricksimonian, line 24: what do oyu think of having the pr body create and update have information on the topic.

api/src/controllers/createTopic.js
```diff
@@ -0,0 +1,26 @@
+import { github } from '../config/index.json';
+import slugify from 'slugify';
+import dotenv from 'dotenv';
+import { githubHandler } from '../utils/githubhandler';
+
+dotenv.config();
```
patricksimonian, line 6: not needed

api/src/controllers/createTopic.js
```diff
@@ -0,0 +1,26 @@
+import { github } from '../config/index.json';
+import slugify from 'slugify';
+import dotenv from 'dotenv';
```
patricksimonian, line 3: not needed

api/src/index.js
```diff
@@ -67,10 +68,12 @@ app.use(cors({ origin: originIsWhitelisted }));
 // authenticated routes
 
 authmware(app);
-// app.use(passport.authenticate('jwt', { session: false }));
+app.use(passport.authenticate('jwt', { session: false }));
 // app.options('/v1/topics', (req, res) => {
```
patricksimonian, line 72: if not used delete

api/src/utils/github.js
```diff
@@ -60,12 +74,55 @@ export const createFile = async (owner, repo, bodyData, ref, topicName, email, n
...
-  const title = `Add new topic ${topicName}`;
-  const body = `Add a new topic to the devhub named ${topicName}`;
+/**
+ *
+ * @param {String} owner
+ * @param {String} repo
+ * @param {String} bodyData
+ * @param {String} ref
+ * @param {String} topicName
+ * @returns {Promise}
```
patricksimonian, line 87: YES!!

api/src/utils/githubhandler.js
```diff
@@ -0,0 +1,77 @@
+import dotenv from 'dotenv';
```
patricksimonian, line 1: dot env is not needed in all files. it only needs to be called once

api/src/utils/githubhandler.js
```diff
@@ -0,0 +1,77 @@
...
+ * @param {String} defaultBranch
+ * @param {String} repo
+ * @param {String} owner
+ * @param {String} req
+ * @param {String} ref
+ * @param {String} bodyData
+ * @param {String} topicName
+ * @returns {Object} {response {status, message, prUrl}}
+ */
+export const githubHandler = async (
```
patricksimonian, line 24: why is this bad?

api/src/utils/githubhandler.js
```diff
@@ -0,0 +1,77 @@
...
+    if (isValidData) {
+      // TODO -- validate topic doesn't already exist
+      // TODO -- validate topic sources are valid
+      // create a git branch on the remote with a naming convention [createTopic/topicname]
+      await createNewRefFromBase(owner, repo, ref);
+      // // create a new file with contents
+      if (operation === 'create') {
+        await createFile(owner, repo, bodyData, ref, topicName);
+      }
+      if (operation === 'update') {
```
patricksimonian, line 48: this is a better case for an `else if` since you are referencing the same variable in the comparison

api/src/utils/githubhandler.js
```diff
@@ -0,0 +1,77 @@
...
+        repo,
+        defaultBranch,
+        topicName,
+        ref,
+      );
+      // URL to the pull Request created ..
+      const pullRequestUrl = pullRequest.data.html_url;
+      response.prUrl = pullRequestUrl;
+      //eslint-disable-next-line
+        console.log(`Pull request for topic ${topicName} created. Check out ${pullRequestUrl} to view the pull request`);
```
patricksimonian, line 65: this is indented to much

api/src/utils/githubhandler.js
```diff
@@ -0,0 +1,77 @@
...
+    } else {
+      response.status = '400';
+      response.statusMessage = 'Bad Request';
+    }
+  } catch (e) {
+    //eslint-disable-next-line
+    console.log("error ->",e);
```
patricksimonian, line 72: e.message

patricksimonian: better of using console.error(e) instead of log

api/src/constants.js
```diff
@@ -7,3 +7,20 @@ export const RESOURCE_VERBS = {
...
+  UPDATE: 'edit topic',
+};
+
+export const PR_TITLE = {
+  CREATE: 'Suggest new topic to the devhub',
+  UPDATE: 'Suggest an edit to topic',
+};
+
+export const PR_BODY = {
+  CREATE: 'Suggest a new topic',
```
patricksimonian, line 24: we talked about this

api/src/controllers/editTopic.js
```diff
@@ -0,0 +1,22 @@
...
+  const { repo, owner, defaultBranch } = github;
+  const bodyData = JSON.stringify(req.body, null, 2);
+  const topicName = slugify(req.body.name.toLowerCase(), '-');
+  const ref = `refs/heads/editTopic/${topicName}`;
+  const response = await createPullRequestFromData(
+    operation,
+    defaultBranch,
+    repo,
+    owner,
+    req,
```
patricksimonian, line 16: The helper function shouldn't need implementation details from the express req.body. This should probably be removed from the function

api/src/utils/githubhandler.js
```diff
@@ -0,0 +1,75 @@
...
+  req,
+  ref,
+  bodyData,
+  topicName,
+) => {
+  let response = { statusMessage: 'Ok', prUrl: '', status: '200' };
+  try {
+    // create new topic and branch
+    // validate topic schema
+    const isValidData = validate(req.body);
```
patricksimonian, line 35: why are we using req body here

api/src/utils/githubhandler.js
```diff
@@ -0,0 +1,75 @@
...
+      await createNewRefFromBase(owner, repo, ref);
+      // // create a new file with contents
+      if (operation === 'create') {
+        await createFile(owner, repo, bodyData, ref, topicName);
+      } else if (operation === 'update') {
+        await updateFile(owner, repo, bodyData, ref, topicName);
+      }
+      // commit  to branch
+      // make pr against ref to base using templates
+      const topicDescription = JSON.parse(bodyData).description;
```
patricksimonian, line 49: and body data here?

api/src/controllers/editTopic.js
```diff
@@ -0,0 +1,22 @@
+import slugify from 'slugify';
+import { github } from '../config/index.json';
+import { createPullRequestFromData } from '../utils/githubhandler';
+
+export const editTopics = async (req, res) => {
+  const operation = 'update';
+  const { repo, owner, defaultBranch } = github;
+  const bodyData = JSON.stringify(req.body, null, 2);
```
patricksimonian, line 8: why is this being stringifies and parsed later on?

api/src/utils/githubhandler.js
```diff
@@ -0,0 +1,75 @@
...
+        operation,
+        owner,
+        repo,
+        defaultBranch,
+        topicName,
+        ref,
+        topicDescription,
+      );
+      // URL to the pull Request created ..
+      const pullRequestUrl = pullRequest.data.html_url;
```
patricksimonian, line 60: this next line could be collapsed into one. This just makes things less verbose / more readable

api/src/constants.js
jas29: The topic description is not a constant, so does not lie in the constants file. It is dynamically loaded, when the pull request is created. I have included it in the createPullRequest function

api/src/utils/githubhandler.js
jas29: using body data to get the topic description, for the pull request body

api/src/utils/githubhandler.js
```diff
@@ -0,0 +1,72 @@
...
+      //eslint-disable-next-line
+      console.log(`Pull request for topic ${topicName} created. Check out ${response.prUrl} to view the pull request`);
+    } else {
+      response.status = '400';
+      response.statusMessage = 'Bad Request';
+    }
+  } catch (e) {
+    //eslint-disable-next-line
+    console.error(e);
+    response.status = e.status;
```
patricksimonian, line 68: i think we want to avoid automatic status codes. You can use a generic 400 status code for now. Sorry i missed this.

patricksimonian: or a 422 better. what sisutaiotns would it fail? only if gh actions fails right?

jas29: I think 442 is a better option, shows up when gh actions fail. In our case that generally happens when the branch already exists or a pull request for a particular topic already exists.

patricksimonian:
I would like to see tests in the next PR

PR #1453, opened 2020-08-20
Code Climate Coverage Docs

jas29:
## Summary
Added a code climate coverage doc, with the link to the code coverage report.
Also includes the Devhub rating badges for the maintainability and test coverage
Added gh action, which uploads test coverage to code climate


.github/workflows/code-climate.yml
```diff
@@ -0,0 +1,85 @@
+name: Generate/Upload all Coverage Reports
+
+on:
+  push:
```
patricksimonian, line 4: maybe lets just do pull events

jas29: sounds good 👍

jas29:
> this looks great but I think before we merge lets just combine this into the previous action that just does node js tests as discussed. No point in running tests twice

Added test coverage process to the unit tests file.

PR #1474, opened 2020-09-22
Suggest an edit to topic getting-started-on-the-devops-platform

patricksimonian:
Suggest new changes to topic getting-started-on-the-devops-platform 
 Resources to help product team become productive as quickly and effectively as possible when using the BC Gov DevOps OpenShift platform.


patricksimonian:
it looks like the edit topic feature doesn't preserve previous configs for sources. We will need to manually address this

PR #1815, opened 2024-02-20
Use main branch for Design System topicRegistry target

ty2k:
## Summary
I've renamed the default branch of the `bcgov/design-system` repository to `main` from `master`, and this PR explicitly updates the design system's `topicRegistry` JSON data to point to that branch.

I think based on [this function](https://github.com/bcgov/devhub-app-web/blob/master/.github/actions/devhub-markdown-frontmatter-verify/src/run.js#L107) in the `devhub-markdown-frontmatter-verify` GitHub Action that this change isn't actually necessary to get the design system content to be picked up by the DevHub build. However, I wanted to make the change proactively in case the build fails, so it's not a bad surprise while you're updating some other part of the application.

### DIFF
--- a/web/topicRegistry/design-system.json
+++ b/web/topicRegistry/design-system.json
@@ -9,6 +9,7 @@
           "url": "https://github.com/bcgov/design-system/",
           "owner": "bcgov",
           "repo": "design-system",
+          "branch": "main",
           "files": [
             "components/about/about.md",
             "components/about/accessibility.md",


MonicaG (commented):
I think your point about [the function](https://github.com/bcgov/devhub-app-web/blob/master/.github/actions/devhub-markdown-frontmatter-verify/src/run.js#L107) is valid. 

My initial reaction is to not add in the 'branch' change as it should pick it up already. The [platform-services-uptime](https://github.com/bcgov/devhub-app-web/blob/43d2fe8d4e41996f87db551248cf1b05f7052a5a/web/topicRegistry/developer-tools.json#L32) repo uses the main branch instead of master. It is [displayed by DevHub](https://developer.gov.bc.ca/Developer-Tools/OpenShift-4-Platform-Services-Reliability-Dashboard) without specifying the branch. 

Although DevHub may need a restart when you rename the branch.

We plan to retire this app in the coming months, so I don't want to add in code if we don't need too.

ty2k:
Thanks for taking a look @MonicaG! I've renamed the branch already and found that doing that broke some inline `<component-preview>` iframe previews on DevHub which were pointing to the `master` branch of `bcgov/design-system`. I hotfixed that by just cutting a new `master` branch from `main` in that repo, and the iframes started working again.

I've updated the design system's internal references to use `main` in [this PR](https://github.com/bcgov/design-system/pull/290/files), but I'll keep that `master` branch going until DevHub is retired.

Like you, I don't want to spend anymore time worrying about this app, so feel free to close this! :)

MonicaG:
Closing without merging as per @ty2k and my above discussion.

